PR #8, opened 2015-09-23
Integrate 3D Secure feature

guzzilar:
## PR Purpose
Add 3D Secure feature into checkout step of WooCommerce checkout page.

#### What's this PR do?
After merge this PR. User will be able to enable the 3D Secure feature in their site from their admin page).

It will change a little bit of the checkout process. From white label to 3D secure means, after you click "checkout button" it will redirect to a page that coming from 3D secure feature. Then, if everything pass. It will be redirected back to a merchant's site for continue a checkout process of WooCommerce.

### DIFF
--- a/omise-api-wrapper.php
+++ b/omise-api-wrapper.php
@@ -20,10 +20,10 @@ if(!class_exists('Omise')){
 		/**
 		 * Retrieve a charge
 		 * @param string $apiKey
-		 * @param Array $chargeInfo
+		 * @param Array $charge_id
 		 * @return mixed
 		 */
-		public static function get_charges($apiKey, $charge_id=""){
+		public static function get_charge($apiKey, $charge_id){
 			$result = self::call_api($apiKey, "GET", "/charges/{$charge_id}");
 			return json_decode($result);
 		}
--- a/omise-api-wrapper.php
+++ b/omise-api-wrapper.php
@@ -16,6 +16,17 @@ if(!class_exists('Omise')){
 			$result = self::call_api($apiKey, "POST", "/charges", $chargeInfo);
 			return json_decode($result);
 		}
+
+		/**
+		 * Retrieve a charge
+		 * @param string $apiKey
+		 * @param Array $chargeInfo
+		 * @return mixed
+		 */
+		public static function get_charges($apiKey, $charge_id=""){
+			$result = self::call_api($apiKey, "GET", "/charges/{$charge_id}");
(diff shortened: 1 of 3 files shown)


omise-wc-gateway.php
```diff
@@ -59,6 +63,12 @@ function init_form_fields() {
 								'label' => __ ( 'Sandbox mode means everything is in TEST mode', $this->gateway_name ),
 								'default' => 'yes' 
 						),
+						'omise_3ds' => array (
+								'title' => __ ( '3DS Support', $this->gateway_name ),
```
fred: title: "3DSecure Support"
label: "Enables 3DSecure on this account"

omise-api-wrapper.php
```diff
@@ -16,6 +16,17 @@ public static function create_charge($apiKey, $chargeInfo){
...
 			return json_decode($result);
 		}
+
+		/**
+		 * Retrieve a charge
+		 * @param string $apiKey
+		 * @param Array $chargeInfo
+		 * @return mixed
+		 */
+		public static function get_charges($apiKey, $charge_id=""){
```
tommy-omise: is there a case where this function returns array of charges?

omise-wc-gateway.php
```diff
@@ -199,10 +210,34 @@ public function process_payment($order_id) {
...
+
+					$result = Omise::create_charge($this->private_key, $data);
+					if ($result->object === "error") {
+						throw new Exception ($result->message);
+					} if ($result->failure_code) {
+						throw new Exception ($result->failure_message);
+					} else {
+						$post_id = wp_insert_post(array(
+							'post_title' 	=> 'Omise Charge Id '.$result->id,
+							'post_type'		=> 'omise_charge_items',
```
tommy-omise: please correct indent here

guzzilar: I'm not sure is it possible to correct an indent.
Wordpress have a standard for php to `Use real tabs and not spaces` but when we use the tab for make an indentation. It will be displayed incorrectly in Github.
ref: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation

Do you have any suggestion about these ?
Thanks krub

omise-api-wrapper.php
guzzilar: Currently, nope 😅.

fred:
:+1:

PR #12, opened 2015-11-15
Added Omise-Version into the cURL request header

guzzilar:
## PR Purpose
- Added Omise-Version into the cURL request header for support the day that the new Omise APIs comes.

(I've already tested on staging. I got `paid` when didn't pass `Omise-Version` and `captured` when pass `Omise-Version: 2014-07-27` properly.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,28 @@
+Change Log
+==========
+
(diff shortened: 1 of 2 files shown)


fred:
:+1:

PR #16, opened 2016-05-22
Support JPY currency 

guzzilar:
#### 1. Objective reason

Supported JPY currency

#### 2. Description of change

1. From the previous version, at the charge process, we always multiply total amount with 100 for `satang`. So, I added 1 condition to check multiply 100 only when `order currency` is `thb`
2. Added `JCB` logo to the Omise payment method.
  screen shot 2559-05-23 at 6 34 35 am

#### 3. Developers affected by the change

`-`

#### 4. Impact of the change

You have to match between your WooCommerce's currency config and Omise account.
i.e. if you want to proceed payment with `JPY` currency.
You have to
1. Set your WooCommerce's currency to JPY (WooCommerce > Settings > General > Currency Options)
2. Use public/secret key from Omise Japan account.

otherwise you will get an error `Payment error:currency is currently not supported`

**noted** You cannot do something like show `JPY` currency sign at the store (product page) but proceed payment with another currency. 

#### 5. Priority of change

:star2: :star2: :star:

#### 6. Alternate solution (if any)**

`-`

### DIFF
--- a/omise-wc-gateway.php
+++ b/omise-wc-gateway.php
@@ -195,12 +195,18 @@ function register_omise_wc_gateway_plugin() {
 						}
 					}
 					
-					$success = false;
+					$success        = false;
+					$order_currency = $order->get_order_currency();
+					if ( 'THB' === strtoupper( $order_currency ) )
+						$amount = $order->get_total() * 100;
+					else
+						$amount = $order->get_total();
+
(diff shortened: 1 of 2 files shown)


guzzilar:
Updated description

guzzilar:
👍

PR #23, opened 2016-07-15
Implement a feature, the configuration for display image of card brands

nimid:
**1. Objective reason**

Accepting card payment for each merchants are not the same. For example, some merchant can accept payments via JCB. Some merchant can not accept payment via JCB but can accept payment via Diners Club.

So it is easier and more comfortable, if the merchant can configure the display card brand logo by themselves.

Related ticket: T373

**2. Description of change**

Add a configuration, Supported card icons, at the back-end, setting page.

screenshot-192 168 99 100 8080 2016-07-15 15-27-19

Add the condition to determine card brand logo display at the front-end, checkout page.

screenshot-192 168 99 100 8080 2016-07-15 15-55-12

All

**4. Impact of the change**

`-`

**5. Priority of change**

Normal

**6. Alternate solution**

`-`

### DIFF
--- a/includes/classes/class-omise-card-image.php
+++ b/includes/classes/class-omise-card-image.php
@@ -124,9 +124,9 @@ if ( ! class_exists( 'Omise_Card_Image' ) ) {
 		public static function is_amex_enabled( $setting ) {
 			if ( isset( $setting['accept_amex'] ) && $setting['accept_amex'] == 'yes' ) {
 				return true;
-			} else {
-				return false;
 			}
+
+			return false;
 		}
 
 		/**
@@ -138,9 +138,9 @@ if ( ! class_exists( 'Omise_Card_Image' ) ) {
 		public static function is_diners_enabled( $setting ) {
 			if ( isset( $setting['accept_diners'] ) && $setting['accept_diners'] == 'yes' ) {
 				return true;
-			} else {
-				return false;
 			}
+
+			return false;
 		}
 
 		/**
@@ -152,9 +152,9 @@ if ( ! class_exists( 'Omise_Card_Image' ) ) {
 		public static function is_jcb_enabled( $setting ) {
 			if ( isset( $setting['accept_jcb'] ) && $setting['accept_jcb'] == 'yes' ) {
 				return true;
-			} else {
-				return false;
 			}
+
+			return false;
 		}
 
 		/**
@@ -166,14 +166,14 @@ if ( ! class_exists( 'Omise_Card_Image' ) ) {
 		public static function is_mastercard_enabled( $setting ) {
 			// Make it backward compatible. If the setting is not configured, the MasterCard logo is display by default.
 			if ( ! isset( $setting['accept_mastercard'] ) ) {
-				return true;
-			} else {
-				if ( $setting['accept_mastercard'] == 'yes' ) {
-					return true;
-				} else {
-					return false;
-				}
+				return self::get_mastercard_default_display();
 			}
+
+			if ( $setting['accept_mastercard'] == 'yes' ) {
+				return true;
+			}
+
+			return false;
 		}
 
 		/**
@@ -185,14 +185,14 @@ if ( ! class_exists( 'Omise_Card_Image' ) ) {
 		public static function is_visa_enabled( $setting ) {
 			// Make it backward compatible. If the setting is not configured, the Visa logo is display by default.
 			if ( ! isset( $setting['accept_visa'] ) ) {
-				return true;
-			} else {
-				if ( $setting['accept_visa'] == 'yes' ) {
-					return true;
-				} else {
-					return false;
-				}
+				return self::get_visa_default_display();
 			}
+
+			if ( $setting['accept_visa'] == 'yes' ) {
+				return true;
+			}
+
+			return false;
 		}
 	}
 }
--- a/includes/classes/class-omise-card-image.php
+++ b/includes/classes/class-omise-card-image.php
@@ -1,0 +1,104 @@
+<?php
+defined( 'ABSPATH' ) or die( "No direct script access allowed." );
+
+if ( ! class_exists( 'Omise_Card_Image' ) ) {
+	class Omise_Card_Image {
+
+		private static function get_html_img($file, $alternate_text) {
+			$url = WC_HTTPS::force_https_url ( WC()->plugin_url() . '/assets/images/icons/credit-cards/' );
+			return "<img src='$url/$file' width='38px' alt='$alternate_text' />";
+		}
+
+		public static function get_amex_default_display() {
+			return 'no';
+		}
+
+		public static function get_amex_image() {
+			return self::get_html_img( 'amex.svg', 'American Express' );
+		}
+
+		public static function get_css() {
+			return 'vertical-align: 5px;';
+		}
+
+		public static function get_diners_default_display() {
+			return 'no';
+		}
+
+		public static function get_diners_image() {
+			return self::get_html_img( 'diners.svg', 'Diners Club' );
+		}
+
+		public static function get_jcb_default_display() {
+			return 'no';
+		}
+
+		public static function get_jcb_image() {
+			return self::get_html_img( 'jcb.svg', 'JCB' );
+		}
+
+		public static function get_mastercard_default_display() {
+			return 'yes';
+		}
+
+		public static function get_mastercard_image() {
+			return self::get_html_img( 'mastercard.svg', 'MasterCard' );
+		}
+
+		public static function get_visa_default_display() {
+			return 'yes';
+		}
+
+		public static function get_visa_image() {
+			return self::get_html_img( 'visa.svg', 'Visa' );
+		}
+
+		public static function is_amex_enabled( $setting ) {
+			if ( isset( $setting['accept_amex'] ) && $setting['accept_amex'] == 'yes' ) {
+				return true;
+			} else {
+				return false;
+			}
+		}
+
+		public static function is_diners_enabled( $setting ) {
+			if ( isset( $setting['accept_diners'] ) && $setting['accept_diners'] == 'yes' ) {
+				return true;
+			} else {
+				return false;
+			}
+		}
+
+		public static function is_jcb_enabled( $setting ) {
+			if ( isset( $setting['accept_jcb'] ) && $setting['accept_jcb'] == 'yes' ) {
+				return true;
+			} else {
+				return false;
+			}
+		}
+
+		public static function is_mastercard_enabled( $setting ) {
+			if ( ! isset( $setting['accept_mastercard'] ) ) {
+				return true;
+			} else {
+				if ( $setting['accept_mastercard'] == 'yes' ) {
+					return true;
+				} else {
+					return false;
+				}
+			}
+		}
+
+		public static function is_visa_enabled( $setting ) {
+			if ( ! isset( $setting['accept_visa'] ) ) {
+				return true;
+			} else {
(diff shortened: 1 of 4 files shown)


includes/classes/class-omise-card-image.php
```diff
@@ -0,0 +1,198 @@
...
+	class Omise_Card_Image {
+		/**
+		 * Compose the given parameters into the string of HTML &lt;img&gt; element
+		 * 
+		 * @param string $file Image file name with extension such as image.jpg
+		 * @param string $alternate_text Alternate text for the image
+		 * @return string HTML &lt;img&gt; element
+		 */
+		private static function get_image( $file, $alternate_text ) {
+			$url = WC_HTTPS::force_https_url( WC()->plugin_url() . '/assets/images/icons/credit-cards/' );
```
guzzilar: @nimid I'm not sure about this one, what if some users still using `http` protocol, is it work for them?

nimid: It can work with HTTP as well.

includes/omise-wc-setting.php
```diff
@@ -0,0 +1,100 @@
...
+
+return array(
+	'enabled' => array(
+		'title'       => __( 'Enable/Disable', $this->gateway_name ),
+		'type'        => 'checkbox',
+		'label'       => __( 'Enable Omise Payment Module.', $this->gateway_name ),
+		'default'     => 'no'
+	),
+	'sandbox' => array(
+		'title'       => __( 'Sandbox', $this->gateway_name ),
```
guzzilar: @nimid ✋ have a question here, would be better to not split this setting array out from the `omise-wc-gateway.php` file since this array still referring some data from `$this`.
what do you think?

nimid: @guzzilar Finally, all of `$this` reference in this file will be removed and it will be changed to use the common function.

Like this, https://github.com/omise/omise-woocommerce/blob/feature/i18n/omise-wc-gateway.php#L47.

guzzilar: @nimid then, how about, just move it back this time that we can make a release for this PR, and implement this approach in that PR instead?

includes/classes/class-omise-card-image.php
```diff
@@ -0,0 +1,198 @@
...
+
+		/**
+		 * Check whether the setting for MasterCard logo is configured and it was set to display or not display
+		 * 
+		 * @param mixed $setting The array that contains key for checking the flag
+		 * @return boolean
+		 */
+		public static function is_mastercard_enabled( $setting ) {
+			// Make it backward compatible. If the setting is not configured, the MasterCard logo is display by default.
+			if ( ! isset( $setting['accept_mastercard'] ) ) {
```
guzzilar: @nimid just share idea that we also write it in another way
```php
public static function is_mastercard_enabled( $setting ) {
    if ( isset( $setting['accept_mastercard'] && $setting['accept_mastercard'] == 'no' ) ) {
        return false;
    }

    // Make it backward compatible. If the setting is not configured or the setting's value is 'yes',
    // the MasterCard logo is display by default.
    return true;
}
```

and we can also use the default setting from `get_*_default_display()` method
i.e.
```php
/**
 * Check whether the setting for American Express logo is configured and it was set to display or not display
 *
 * @param mixed $setting The array that contains key for checking the flag
 * @return boolean
 */
public static function is_amex_enabled( $setting ) {
    $setting = isset($setting['accept_amex']) ? $setting['accept_amex'] : self::get_amex_default_display();

    if ($setting == 'yes') {
        return true;
    }

    return false;
}
```

Btw, I'm ok both na 😄

nimid: Yes, I got your idea.

For me, there are not much different. Both designs provide the same result.

guzzilar: For me, it's different in terms of code clean and reduce some nested `if ... else`.
this one also
```php
public static function is_jcb_enabled( $setting ) {
    if ( isset( $setting['accept_jcb'] ) && $setting['accept_jcb'] == 'yes' ) {
        return true;
    } else {
        return false;
    }
}
```

it could be
```php
public static function is_jcb_enabled( $setting ) {
    if ( isset( $setting['accept_jcb'] ) && $setting['accept_jcb'] == 'yes' ) {
        return true;
    }

    return false;
}
```

But yeah, it provides the same result and not much different as you said 😂

👍

nimid: Yes.

robinclart: I think it will make more sense to increase clarity and remove as much conditional branching as possible.

robinclart: Especially when two of the branch could be merged into one.

includes/omise-wc-setting.php
```diff
@@ -0,0 +1,100 @@
+<?php
+defined( 'ABSPATH' ) or die( "No direct script access allowed." );
+
+return array(
```
guzzilar: @nimid from the commit that you removed text domain out from the `__()` method, https://github.com/omise/omise-woocommerce/pull/23/commits/4e6b1ea3bcedcbec6ddc8931afdf7a5adfe007f6

I saw article that mentions about text domain here https://codex.wordpress.org/I18n_for_WordPress_Developers#Text_Domains

Would be better to not remove `text domain` here, but, change to some variable instead?
something like,
```php
$text_domain = 'Omise';

__( 'Enable/Disable', $text_domain );
```
(at least for this time until we merge #24 )?

nimid: If we define the text domain in this file, it means we need to make it consistent with all files that contain the static message. Even define a text domain in this file for this merge but the next coming merge, we need to refactor it again.

For this time I will remove all translation function from this file to make it simpler and prevent misleading.

I saw that article and pay attention to it as well. As can be seen from I have developed [a common function to be reused to translate the message](https://github.com/omise/omise-woocommerce/blob/feature/i18n/omise-util.php#L84).

At the beginning of this feature development, I already know that the translation will be developed or merged with another branch, feature/i18n, so I decided to focus on the scope of the feature. For the other stuffs, I still keep it intact like before I contribute with this project, if it not break, conflict, or affect the system to not working.

So one of benefits of this decision is to reduce the scope of review and testing time and cost.

However, thank you for your mention on the bad old code.

guzzilar: @nimid for the reason that I asked you to complete the features within this PR (without waiting for another PR to be merged) because I'm not sure how long we gonna take for code review in another PR.

Well, the benefit is, we can merge this pr to `master` and make a new release without needed to wait for maybe 3 - 5 days later to complete everything since those 2 prs (#23 & #24) didn't related with each other at all (we can have card brand logo settings without translation). 
So, We can ship it one by one.

For the text domain, from the reason above. I asked you because the previous code we had defined the text domain `Omise`. 
So, if we gonna ship it one by one, we should make it backward compatible as much as possible with the previous version instead of waiting for another future PR to complete the code.

(and yep, I saw that PR #24 already)

nimid: Yes.

guzzilar: @nimid so, I still prefer to implement with the text domain as below
```php
$text_domain = 'Omise';

__( 'Enable/Disable', $text_domain );
```

Because we can't guarantee that no one (developers) who's downloaded our plugin will not implement `I18n` feature with the text domain `Omise` by themselves, so if we remove `__()` method it may break some website that they already implemented `I18n`.

nimid: I agree with backward compatibility but what is impact (break website) you mentioned?

guzzilar: translation that related with `Omise` text domain?

nimid: Yes.

guzzilar: @nimid I'm still thinking that what's gonna happen if we remove `Omise` text-domain and someone already downloaded our plugin and did implement I18N feature by themselves,

their implementation'd break because their translation can't refer to our previous text-domain?

nimid: If they translate or develop anything by themselves, it means that they response and manage it by themselves on their own way, on their own branch, on their own private.

This project is an open-source. They can contribute or show what they translate or develop publicly on GitHub or not on GitHub that is very nice and valuable and we will support it.

If we found that any our customers even one have done that (private or public), the concern that you worry will be considered to be benefit of customers.

Note the reason I decided to develop in that way
During development, I have also considered the old code in part of translation. I found many conflicts and inconsistencies.

Some message has not been implemented the translation even it is a message in context of Omise.
Some message has been implemented the translation but it has not been defined the text domain.
Some message has been implemented the translation but it has been defined the text domain, woothemes or woocommerce.

It does not make sense and it should be refactored.

Including, some message has been translatable but it has no translation file (PO or POT file). This lack of translation file makes me understand that the translation feature is not completed and not finished yet. So if we decide to make it, we should make it done without conflicts.

If we have not plan to develop the translation or the translation feature will be not release soon. I agree to remain the text domain.

Now, I feel the review process is too long and it should be enough. From my experience, this stuff can be done in review process not more than a few days even it has long discussion or it has many concerns.

Conclusion
I will bring the text domain back because I do not want anymore delay and I want to release this feature for customers to use.

guzzilar: @nimid agreed with above, except `If we have not plan to develop the translation or the translation feature will be not release soon. I agree to remain the text domain.`

like I said before, I'm not sure how long we gonna spend for another review and I also want to release this version without have to wait for another review.

So, if you're agree for `remain the text domain If we have not plan to develop the translation or the translation feature will be not release soon.` then, I'll wait for another commit and merge it asap.

thanks :)

includes/omise-wc-setting.php
```diff
@@ -0,0 +1,102 @@
...
+		'description' => __( 'This controls the title which the user sees during checkout.', $text_domain ),
+		'default'     => __( 'Omise Payment Gateway', $text_domain )
+	),
+	'payment_action' => array(
+		'title'       => __( 'Payment Action', $text_domain ),
+		'type'        => 'select',
+		'description' => __( 'Manual Capture or Capture Automatically', $text_domain ),
+		'default'     => 'auto_capture',
+		'class'       => 'wc-enhanced-select',
+		'options'     => $this->form_field_payment_actions(),
```
guzzilar: @nimid still have `$this` reference here,
from 
```php
public function form_field_payment_actions() {
    return array(
        'auto_capture'   => __( "Auto Capture", $this->gateway_name ),
        'manual_capture' => __( "Manual Capture", $this->gateway_name )
    );
}
```

can you change to
```php
'payment_action' => array(
    '...'         => '...',
    'options'     => array(
        'auto_capture'   => __( "Auto Capture", $this->gateway_name ),
        'manual_capture' => __( "Manual Capture", $this->gateway_name )
    ),
    '...'         => '...'
),
```
instead

guzzilar:
@nimid pls review the commit https://github.com/omise/omise-woocommerce/pull/23/commits/dd397bbd6e3ac14cae2f9af1d8f2467b6c82a460

thanks

guzzilar:
@nimid are you ok with https://github.com/omise/omise-woocommerce/pull/23/commits/dd397bbd6e3ac14cae2f9af1d8f2467b6c82a460 ?

I'm done with it, will merge after your review.

nimid:
Yes. It is OK. I have merged it (https://github.com/omise/omise-woocommerce/pull/23/commits/b804659ab2a531c1ab8b2924ebabbd0bb1c351ad).

guzzilar:
👍

PR #26, opened 2016-08-10
Update supported currencies

AsukaOmise:
**1. Objective reason**

To update merchants about the supported countries

**2. Description of change**

Add Japan to Readme

None

**4. Impact of the change**

None

**5. Priority of change**

Normal

**6. Alternate solution (if any)**

None


nimid:
Thank you @AsukaOmise.

PR #27, opened 2016-08-24
Use an inline style for card brand image

nimid:
**1. Objective reason**

Define the inline style for the card brand image.

When the customer has define the rule (cascading style sheet) for the image size by themselves or by theme that they use and that rule impacts to the display of the card brand image, the display of image will not proportional.

The image below shows an example of the display of card brand image is not proportional.

screenshot-192 168 99 100 8080 2016-08-24 22-10-38

Related ticket: T914

**2. Description of change**

Remove an attribute, width, from the image and add an inline style to override the user's style.

A reference for the cascading order: [W3C The cascade](https://www.w3.org/TR/CSS22/cascade.html#cascade)

All

**4. Impact of the change**

`-`

**5. Priority of change**

Normal

**6. Alternate solution**

Bundle the card brand images instead of using the images of WooCommerce and define the cascading style sheet (CSS) to control the display and override the other CSS.

### DIFF
--- a/includes/classes/class-omise-card-image.php
+++ b/includes/classes/class-omise-card-image.php
@@ -12,7 +12,7 @@ if ( ! class_exists( 'Omise_Card_Image' ) ) {
 		 */
 		private static function get_image( $file, $alternate_text ) {
 			$url = WC_HTTPS::force_https_url( WC()->plugin_url() . '/assets/images/icons/credit-cards/' );
-			return "<img src='$url/$file' style='width: 38px;' alt='$alternate_text' />";
+			return "<img src='$url/$file' class='Omise-CardBrandImage' style='width: 38px;' alt='$alternate_text' />";
 		}
 
 		/**
--- a/includes/classes/class-omise-card-image.php
+++ b/includes/classes/class-omise-card-image.php
@@ -12,7 +12,7 @@ if ( ! class_exists( 'Omise_Card_Image' ) ) {
 		 */
 		private static function get_image( $file, $alternate_text ) {
 			$url = WC_HTTPS::force_https_url( WC()->plugin_url() . '/assets/images/icons/credit-cards/' );
-			return "<img src='$url/$file' width='38px' alt='$alternate_text' />";
+			return "<img src='$url/$file' style='width: 38px;' alt='$alternate_text' />";
 		}
 
 		/**


includes/classes/class-omise-card-image.php
```diff
@@ -12,7 +12,7 @@ class Omise_Card_Image {
 		 */
 		private static function get_image( $file, $alternate_text ) {
 			$url = WC_HTTPS::force_https_url( WC()->plugin_url() . '/assets/images/icons/credit-cards/' );
-			return "<img src='$url/$file' width='38px' alt='$alternate_text' />";
+			return "<img src='$url/$file' style='width: 38px;' alt='$alternate_text' />";
```
guzzilar: @nimid ✋ have one question here, do you have any idea/plan to support in case people want to overwrite this rule (width 38px) by themselves?
For example, they want to control the sizing of the icons on desktop/mobile screens to have difference width.

nimid:
Tested on:

- Safari 9.1.2

screen shot 2016-08-24 at 11 21 27 pm

- Firefox 48.0.1

screen shot 2016-08-24 at 11 23 30 pm

- Google Chrome 52.0

screen shot 2016-08-24 at 11 18 29 pm

- Opera 39.0

screen shot 2016-08-24 at 11 18 51 pm

includes/classes/class-omise-card-image.php
nimid: The customer can create style to override this rule or modify this rule by themselves. To develop the system, we can develop the system provide the size customization.

What do you think?

guzzilar: I think just provide a class attr for this element and set the default rule to width 38px, class attr just for customer who may create style to override the rule as you mentioned

nimid: Yes, but that solution can not solve the problem for a present customer on the related ticket.

And how about the non-technical customer that they can not, do not know or do not want to modify the style sheet?

We provide the best image size and those images are SVG file the responsive has enabled normally.

guzzilar: @nimid ah, I meant, from this line,
```php
return "<img src='$url/$file' style='width: 38px;' alt='$alternate_text' />";
```
just add more class attr here
```php
return "<img src='$url/$file' class='OmiseCardBrandIcons' style='width: 38px;' alt='$alternate_text' />";
```

that customer can use `.OmiseCardBrandIcons` to override `style='width: 38px;'`,

---
alternate way,
have another css file to set 
```css
.OmiseCardBrandIcons {
  width: 38px;
}
```
instead of set style inline.

nimid: @guzzilar Did you tested it? Please show me if it work.

The weight of class style sheet is lower than weight of inline sheet. Check the reference that I provied in the pull request description.

nimid: @guzzilar I agree with the solution to provide a style class for the image. It deliver the benefit to customer. But the redundant configuration is unnecessary.

If the customer really want to customize, they can customize it by using that solution or by using so many other flexible CSS implementation style.

nimid: @guzzilar What do you think?

guzzilar: @nimid sry, not sure I'm on the same page as you 😅 

From what I think is, right now you have
```php
return "<img src='$url/$file' style='width: 38px;' alt='$alternate_text' />";
```

just add another attribute, `class='OmiseCardBrandIcons'` there

So, it's gonna be,
```php
return "<img src='$url/$file' class='OmiseCardBrandIcons' style='width: 38px;' alt='$alternate_text' />";
```

and whenever (maybe rare case) customer want to override it, they can do
```css
.OmiseCardBrandIcons {
  width: 80px !important;
}
```

`!important` may not be the good choice, but at least, it allows customer to override `inline stylesheet` in case they want to, without touch our plugin files.

guzzilar: and if no one override it, your inline stylesheet still works there as well.

nimid: @guzzilar I agree with the last solution, define the class name without style to provide the customer create their own style to override the display.

guzzilar: 👍

guzzilar:
👍

PR #28, opened 2016-08-25
Remove unused unit tests of omise php

nimid:
**1. Objective reason**

Solve the problem when the user install plugin via WordPress back-end on the server that run on Microsoft Windows. The problem is user found the error about unable to copy the long file and folder name.

Related ticket: T919

**2. Description of change**

- Remove an unused folder and files that related with unit testing of library, omise-php.

Some files of it cause the error. Unit testing is not used to run or function in any part. Some files have long file name and it stored in the long directory name also.

For an example:
`./includes/libraries/omise-php/tests/fixtures/api.omise.co/customers/cust_test_4zmrjg2hct06ybwobqc/cards/card_test_4zmrjfzf0spz3mh63cs-delete.json`

Count the characters of the above example including the slash mark is 146 characters.

So this long path plus the user directory can exceed the limit that Microsoft Windows can accept, 260 characters.

Note:
To remove the unused files and folder, it was useful in reducing the time to download and upload plugin as well.

A reference for [the maximum path length limitation of Microsoft Windows](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx).

The user that install the plugin on Microsoft Windows.

**4. Impact of the change**

`-`

**5. Priority of change**

Normal

**6. Alternate solution**

`-`


guzzilar:
@nimid 👍 
also, I have an idea that after we can upload this plugin to WooCommerce/WordPress stores, we can also remove `omise-php` lib from here and implement it as `git submodule` instead to make it easy when we have update `omise-php` lib (don't need to go to every plugin and do copy/paste, just run git submodule to install).

guzzilar:
👍

PR #36, opened 2017-06-14
Give a proper deigned-structure for payment methods

guzzilar:
> This PR is a part of code refactoring, and OmiseWooComemrce v2.0

### 1. Objective

With Internet Banking feature, it would create a tons of duplicated code without refactoring the current code (since the original structure didn't design to support multiple payment methods in one plugin).

This PR aims to redesign the payment method structure, make it more extensible and move some common methods into a delegated-payment-class. 

Related ticket: 🙅

### 2. Description of change

**2.1 Introducing a new plugin structure.**
Previously we keep `omise-wc-gateway.php` at root folder of the plugin.
The question is, what if we have to implement more payment methods (i.e. Internet Banking, Alipay, etc.)? Where those classes should go, and how to maintain tons of common methods code.

The solution for this PR is, having a proper structure for those payment method classes, which is,
```
./includes/gateway
./includes/gateway/class-omise-payment.php // Base payment class that extends WC_Payment_Gateway
./includes/gateway/class-omise-payment-creditcard.php
./includes/gateway/class-omise-payment-internetbanking.php
./includes/gateway/class-omise-payment-alipay.php

.. and so on.
```

**2.2 Bring related payment gateway methods back to `Omise_Payment_Creditcard`.**
From the payment method perspective, rather than having different class/ place to handle a payment action. I decided to have methods (function) inside a payment method class to perform each of actions.

Think it as a processor, what does `Omise_Payment_Creditcard` can do? 'manual capture', '3-D Secure callback', 'authorize', 'capture' and so on. Those actions will be implemented inside a `Payment_Creditcard` only since it's behaviors of Credit Card.

From the above concept, I moved some methods from `Omise_Hook` back into `Omise_Payment_Creditcard`.

**2.3 Refactoring code & clean up**
Code refactoring, clean some of long-code-method into small chunk, make it better to maintenance.

**2.4 Setting menu's url has been changed to `wc-settings&tab=checkout&section=omise` (previously, `section=wc_gateway_omise`)**

before:  
http://127.0.0.1/wp-admin/admin.php?page=wc-settings&tab=checkout&section=wc_gateway_omise

after:  
http://127.0.0.1/wp-admin/admin.php?page=wc-settings&tab=checkout&section=omise

**2.5 Enhance 'order note' in most of payment steps.**
 
#### 3. Quality assurance

**🔧 Environments:**

- **Platform version**: WooCommerce v3.0.8
- **PHP version**: 7.0.16.

**✏️ Details:**

There are various cases that need to be tested, please check from the test cases below.

**3.1. ✅ Charge (auto capture) with none 3-D Secure payment would be successful as normal.**

**3.2. ✅ Charge (authorize only) with none 3-D Secure payment would be successful as normal.**

**3.3. ✅ Payment will be failed and raise an error message properly when make a charge using [failed charge card](https://www.omise.co/api-testing).**

**3.4. ✅ Charge (auto capture) with 3-D Secure payment would be successful as normal.**

**3.5. ✅ Charge (authorize only) with 3-D Secure payment would be successful as normal.**

**3.6. ✅ Be able to perform 'manual capture' action through WooCommerce's order detail page.**

**3.7. ✅ Be able to save card with `customer` API.**

#### 4. Impact of the change

Some part of localization will be broken.
I will check & solve it in another refacotring PR.

#### 5. Priority of change

Normal

#### 6. Additional Notes

No


guzzilar:
👍

PR #41, opened 2017-06-28
Support Internet Banking payment method.

guzzilar:
### 1. Objective

To add support [Omise's Internet Banking feature](https://www.omise.co/offsite-payment).

### 2. Description of change

2.1. New payment method code `omise_internetbanking`.
2.2. New callback url `?wc-api=omise_internetbanking_callback&order_id=xx`
2.3. New setting tab at admin, WooCommerce Checkout page.
2.4. New payment method at the checkout page (Internet Banking).

#### 3. Quality assurance

- **Platform version**: WooCommerce v3.0.9
- **PHP version**: 7.0.16.

**✏️ Details:**

**3.1. Once you install the plugin and enable it. You must see this setting in the WooCommerce > Setting > Checkout page.**

screencapture-127-0-0-1-wp-admin-admin-php-1498679181346

**3.2. After you enable the payment method. Now the Internet Banking payment method will be appeared in the checkout page.**

screen shot 2560-06-29 at 2 47 43 am

**3.3. Test checkout with internet banking payment method and mark a payment as `failure`**
screen shot 2560-06-29 at 2 58 40 am

You will be redirected back to the checkout page with an error message on a screen

screen shot 2560-06-29 at 2 58 22 am

_Omise Dashboard, charge detail page._
screen shot 2560-06-29 at 3 25 29 am

**3.4. Test checkout with internet banking payment method and mark a payment as `success`.**

screen shot 2560-06-29 at 2 58 42 am

You will be redirected back to the `order-received` page.
screen shot 2560-06-29 at 3 04 37 am

_Omise Dashboard, charge detail page._
screen shot 2560-06-29 at 3 25 05 am

**3.5. Check at admin, WooCommerce order detail page , you will see new order was made by using  `internet banking` as a payment method.**

screen shot 2560-06-29 at 3 21 32 am

#### 4. Impact of the change

No.

#### 5. Priority of change

Normal

#### 6. Additional Notes

- You will need to assign your public key & secret key again in another payment method. (now 2 payment methods cannot share the same key).
- Which means, you may need to **reset your secret key at Omise dashboard** to get a `live secret key` that you can use with Internet Banking payment.

PR #46, opened 2017-07-10
Now merchant doesn't need to re-set Omise keys in every new payment methods.

guzzilar:
#### 1. Objective

Previously merchant needs to set Omise public and secret keys twice on both Credit Card and Internet Banking configuration pages, which creates a bad user experience (as described in PR #41, section number 6).

**Related information**:
Related issue(s): No

#### 2. Description of change

- Now merchant doesn't need to set any keys on a new payment method. It would automatically use the same keys as from Credit Card configuration page.

#### 3. Quality assurance

**🔧 Environments:**

- WordPress v4.8
- WooCommerce v3.1.0
- PHP v5.6.30

**✏️ Details:**

✅ **3.1. Test create charge with the Internet Banking payment method.**
_**Expectation**_ Buyer must be able to create a new charge right away.

✅ **3.2. Now, try to set a wrong key at Credit Card configuration page and make a new charge with Internet Banking again.**
_**Expectation**_ Charge must be failed because the key in Credit Card configuration page has changed (to prove that right now, Internet Banking payment method use the same key as set in Credit Card payment method).

#### 4. Impact of the change

No

#### 5. Priority of change

Normal

#### 6. Additional Notes

No.

### DIFF
--- a/includes/gateway/class-omise-payment-creditcard.php
+++ b/includes/gateway/class-omise-payment-creditcard.php
@@ -43,99 +43,80 @@ function register_omise_creditcard() {
 		 * @see woocommerce/includes/abstracts/abstract-wc-settings-api.php
 		 */
 		function init_form_fields() {
-			$this->form_fields = array(
-				'enabled' => array(
-					'title'       => __( 'Enable/Disable', 'omise' ),
-					'type'        => 'checkbox',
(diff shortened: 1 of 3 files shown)


guzzilar:
👍

PR #49, opened 2017-07-13
Humanize messages that will be displayed on a user's screen

guzzilar:
### 1. Objective

Because currently messages in the plugin kinda robot-ish, it can tell what is going on, but not enough to helping user to diagnose issues that has been raised on their screen by themselves.

For example, a payment result validation at the 3-D Secure callback function.
```
Order not found: Your card might be charged already.
Please contact our support team if you have any questions
```
This one is almost good, but can be better. 'Order not found'.
As a programmatic, it does make senses because this error will be raised if a method cannot retrieve an order object from a given `order id`.

But as a user, it tells nothing except something is wrong.
We can humanize it more by the following message.

```
We cannot validate your payment result:
Note that your payment might already has been processed.
Please contact our support team if you have any questions
```

### 2. Description of change

2.1. Enhance all messages in the plugin

2.2. Translate all of messages into JA, ID.

2.3. Add Grunt.js task. (now you can use 'grunt makepot' to build a 'pot' file.)

#### 3. Quality assurance

- **Platform version**: WooCommerce v3.1.1
- **PHP version**: 5.6.30

**✏️ Details:**

It's mostly about word, sentence, message. 
In order to check, you can look at https://github.com/omise/omise-woocommerce/blob/a332173499d4f098dea0cb8068366de8a5de8a57/languages/omise-ja.po

#### 4. Impact of the change

No.

#### 5. Priority of change

Normal

#### 6. Additional Notes

There are some parts that haven't been translated left but it's my intention. I want to test translate with GlotPress.


guzzilar:
👍

PR #59, opened 2017-08-11
Support WooCommerce 2.x series & PHP 5.4

guzzilar:
#### 1. Objective

There are bunch of internal support tickets asking for make Omise-WooCommerce support for WooCommerce v2.x series (and also asking to solve an issue after upgrade plugin on WooCommerce v2.x) 

#### 2. Description of change

- So I went to check and found that there are some methods that has been introduced only in WooCommerce v3.x series. But not that much to impact code structure, so I decided to add some condition to check if user is still using WooCommerce legacy version.

- Also, update PHP syntax to support PHP v5.4.x

#### 3. Quality assurance

**🔧 Environments:**

- **WordPress**: `v4.8.1`.
- **WooCommerce**:  `v3.1.1`.
- **PHP**: `v5.4.45` / `v5.6.30`

**✏️ Details:**

1) Make sure that you can create a new charge on both WooCommerce v2.x and v3.x with Credit Card / Alipay / Internet Banking payment methods

2) Make sure that you can perform `manual sync` action at order detail page.

3) Make sure that you can attach a `charge id` into an order object.

#### 4. Impact of the change

No

#### 5. Priority of change

Normal

#### 6. Additional Notes

- PHP 5.4.x was in [End of Life](http://php.net/eol.php) state since almost 2 years. We strongly suggest that merchants should upgrade their server to at least PHP v5.6 or above.

### DIFF
--- a/includes/gateway/class-omise-payment-alipay.php
+++ b/includes/gateway/class-omise-payment-alipay.php
@@ -102,9 +102,15 @@ function register_omise_alipay() {
 					case 'pending':
 						$this->attach_charge_id_to_order( $charge['id'] );
 
-						$order->set_transaction_id( $charge['id'] );
 						$order->add_order_note( sprintf( __( 'Omise: Redirecting buyer out to %s', 'omise' ), esc_url( $charge['authorize_uri'] ) ) );
-						$order->save();
+
+						/** backward compatible with WooCommerce v2.x series **/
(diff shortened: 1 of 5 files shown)


guzzilar:
👍

guzzilar:
👍

PR #61, opened 2017-09-07
Add Omise Setting page (separate between general settings and payment method settings)

guzzilar:
#### 1. Objective

Previously, Omise configs were located under the `WooCommerce > Checkout > Omise Credit / Debit Card` page, which is in fact, these configs will be used for other payment methods as well.
Without having its own separated page could confusing user on the setting (i.e. user sets pkey/skey at Credit Card payment method but can also use for IB and Alipay).

Also, Webhook feature is coming. Means we need to have some place to have a webhook url setting and it shouldn't be under the `WooCommerce > Checkout > Omise Credit / Debit Card` or other payment method pages.

**Related information**:
Related issue(s): 🙅

#### 2. Description of change

1. Setting page under the Omise menu at WordPress admin page.
    screencapture-127-0-0-1-wp-admin-admin-php-1504799768907

2. Removed `Payment Settings` section from `WooCommerce > Checkout > Omise Credit / Debit Card` page.
    screencapture-127-0-0-1-wp-admin-admin-php-1504800149487

#### 3. Quality assurance

**🔧 Environments:**

- **WordPress**: 3.8.1
- **WooCommerce**: 3.1.2.
- **PHP**: 5.6.30

**✏️ Details:**

1. **Make sure that this implementation won't cause any trouble with the current settings that has been set at the previous setting page.**

    1.1 Assume that a user already installed Omise and config his public/ secret keys at his store (if not, then install Omise-WooComemrce v3.0 and setup first).

    1.2. Deploy this PR.

    1.3. Then go to Omise new setting page, **your current settings should be displayed there properly**.

2. **Make sure that your store can perform checkout with credit card and other payment methods properly**

    2.1 Continue from [1]. Without any extra steps, your store should still work as normal.

3. **Make sure that user can setup new Omise-WooCommerce plugin and able to config his Omise settings properly.**

    3.1 Assume that user just about to installs Omise for the first time (if not, then re-install WordPree & WooCommerce & Omise from scratch).

    3.2 Test config public / secret keys and enable test mode.

    3.3 Your config should be persisted there.

4. **Make sure that you can use your keys from step [3] to make charge with any payment methods (Credit / Debit Card, Alipay, Internet Banking).**

#### 4. Impact of the change

No

#### 5. Priority of change

Normal

#### 6. Additional Notes

No

### DIFF
--- a/includes/admin/class-omise-page-settings.php
+++ b/includes/admin/class-omise-page-settings.php
@@ -7,11 +7,39 @@ if ( class_exists( 'Omise_Page_Settings' ) ) {
 }
 
 class Omise_Page_Settings {
+	/**
+	 * @var Omise_Setting
+	 */
 	protected $settings;
 
+	/**
+	 * @since 3.1
+	 */
 	public function __construct() {
 		$this->settings = new Omise_Setting;
 	}
+
+	/**
+	 * @return array
+	 *
+	 * @since  3.1
+	 */
+	protected function get_settings() {
+		return $this->settings->get_settings();
+	}
+
+	/**
+	 * @param array $data
+	 *
+	 * @since  3.1
+	 */
+	protected function save( $data ) {
+		$this->settings->update_settings( $data );
+	}
+
+	/**
+	 * @since  3.1
+	 */
 	public static function render() {
 		global $title;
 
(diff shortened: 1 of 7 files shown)


guzzilar:
👍

PR #70, opened 2017-09-22
WIP : Introduce Chatbot feature!

guzzilar:
#### 1. Objective

🎵 Communication on social network is like a busy crossroad. People come to you in all directions at different degrees of urgency, and most of the time, it’s the same old questions. Keeping up with your daily tasks while running an online store isn’t all that easy, but we're here to help!

**Introducing new feature, Chatbot, your round-the-clock store assistant on Facebook Messenger.**

Why keep your customers waiting for a simple answer to a simple question? With this new version of the Omise plugin for website on WooCommerce, you can kick back and relax while the Chatbot takes care of all the simple conversations through Messenger. This feature is automatically enabled with effortless implementation. Chatbot is the perfect fit for a website on WooCommerce with a Facebook page, as it will retrieve your product’s information from the website and link it on Messenger.

_basically, copied from https://www.omise.co/omise-woocommerce-chatbot lol_

#### 2. Description of change

> ⚠️ Note that this is just a container PR and it doesn't has any implementation in its PR.

Because this is such a huge implementation so, to understand all of the changes, you are going to dig into each of sub-task PRs as follows:

1. ✅ **#67** : Proper display a setting page and able to save the settings
2. ✅ **#68** : Be able to register Omise Chatbot feature into Facebook Messenger application
3. ✅ **#69** : Setup greeting screen and be able to retrieve Facebook payload and response back
4. ✅ **#71** : Preparing Facebook Messenger template file structure
5. ✅ **#73** : Display featured product list
5. ✅ **#74** : Display product category list

> _Installation Instruction and screenshot are coming._

**In progress**
- ⬜️ Display gallery of a product
- ⬜️ Display product list
- ⬜️ Be able to purchase an item from Facebook Messenger
- ⬜️ Be able to check an order status
- ⬜️ Add 'Persistent Menu' to guide user on basic actions they can perform on Chatbot.

#### 3. Quality assurance

Same as section [2] Description of change. You are going to check each of sub-task PRs one by one.

#### 4. Impact of the change

No

#### 5. Priority of change

Normal

#### 6. Additional Notes

Enjoy.

PR #71, opened 2017-09-25
Chatbot: Preparing Facebook Messenger template file structure

guzzilar:
> ⚠️ This PR is a sub-task of Chatbot feature (PR #70)

#### 1. Objective

If we take a look at this document https://developers.facebook.com/docs/messenger-platform/reference/send-api. Facebook has quite complicated (and rich) template structure that allows 3rd-party developer to integrate and customize a message output that Bot will use to reply to a user.

Because of that, we need to carefully implement this part, otherwise it would easily turns to [Spaghetti Code](https://en.wikipedia.org/wiki/Spaghetti_code).
So this PR aims to make clear on a structure that we would go with for the Facebook Messenger Templates _(a bit irony that to solve a complex structure, this PR quite also complex.. any thoughts are welcome!)_.

#### 2. Description of change

- Implement with 2 templates and 1 plain-text content. (generic template and button template)
- Mock an output that we can see how it is going to be used.
    screen shot 2560-09-25 at 7 34 26 pm

#### 3. Quality assurance

**🔧 Environments:**

- **PHP**: v5.4.45
- **WordPress**: v4.8.2
- **WooCommerce**: v3.1.2

**✏️ Details:**

1. Make sure that these templates can be used by Chatbot properly. (see screenshot at section 2)

#### 4. Impact of the change

No

#### 5. Priority of change

Normal

#### 6. Additional Notes

- An output messages are mock-data. It would be changed in another PR.

- To help demonstrate how these classes related to each others.
Here are outputs of each class when you try initiate its classes.
```php
print_r((new Omise_Chatbot_Component_Attachment)->to_array());
// output
Array(
    [attachment] => Array(
        [type] => 
        [payload] => Array()
    )
)

// Omise_Chatbot_Component_Template extends from Omise_Chatbot_Component_Attachment
print_r((new Omise_Chatbot_Component_Template)->to_array());
// output
Array(
    [attachment] => Array(
        [type] => template
        [payload] => Array(
            [template_type] =>
        )
    )
)

// Omise_Chatbot_Component_Template_Button extends from Omise_Chatbot_Component_Template
print_r((new Omise_Chatbot_Component_Template_Button)->to_array());
// output
Array(
    [attachment] => Array(
        [type] => template
        [payload] => Array(
            [template_type] => button
            [text] => 
            [buttons] => Array()
        )
    )
)
```

### DIFF
--- a/includes/chatbot/components/buttons/class-omise-chatbot-component-button-orderstatus.php
+++ b/includes/chatbot/components/buttons/class-omise-chatbot-component-button-orderstatus.php
@@ -6,10 +6,6 @@ if ( class_exists( 'Omise_Chatbot_Component_Button_Orderstatus' ) ) {
 }
 
 class Omise_Chatbot_Component_Button_Orderstatus extends Omise_Chatbot_Component_Button_Postback {
-	/**
-	 * @param string $title
-	 * @param string $payload
-	 */
 	public function __construct() {
 		$this->attributes['title']   = __( 'Check order status', 'omise' );
 		$this->attributes['payload'] = Omise_Chatbot_Payloads::ACTION_ORDER_STATUS;
--- a/includes/chatbot/components/buttons/class-omise-chatbot-component-button-orderstatus.php
+++ b/includes/chatbot/components/buttons/class-omise-chatbot-component-button-orderstatus.php
@@ -1,0 +1,17 @@
+<?php
+defined( 'ABSPATH' ) or die( 'No direct script access allowed.' );
+
+if ( class_exists( 'Omise_Chatbot_Component_Button_Orderstatus' ) ) {
+	return;
+}
+
+class Omise_Chatbot_Component_Button_Orderstatus extends Omise_Chatbot_Component_Button_Postback {
+	/**
+	 * @param string $title
+	 * @param string $payload
+	 */
+	public function __construct() {
+		$this->attributes['title']   = __( 'Check order status', 'omise' );
+		$this->attributes['payload'] = Omise_Chatbot_Payloads::ACTION_ORDER_STATUS;
+	}
+}
--- a/includes/chatbot/class-omise-chatbot-client.php
+++ b/includes/chatbot/class-omise-chatbot-client.php
@@ -1,0 +1,62 @@
+<?php
+defined( 'ABSPATH' ) or die( 'No direct script access allowed.' );
+
+if ( class_exists( 'Omise_Chatbot_Client' ) ) {
+	return;
+}
+
+class Omise_Chatbot_Client {
+	/**
+	 * @param  string $url
+	 * @param  mixed  $body
+	 *
+	 * @return WP_Error|array  The response or WP_Error on failure.
+	 *
+	 * @since  3.2
+	 */
+	public function get( $url, $body = null ) {
+		return wp_safe_remote_get(
+			$url,
+			array(
+				'timeout' => 60,
+				'body'    => $body
+			) 
+		);
+	}
+
+	/**
+	 * @param  string $url
+	 * @param  mixed  $body
+	 *
+	 * @return WP_Error|array  The response or WP_Error on failure.
+	 *
+	 * @since  3.2
+	 */
+	public function post( $url, $body = null ) {
+		return wp_safe_remote_post(
+			$url,
+			array(
+				'timeout' => 60,
+				'body'    => $body
+			) 
+		);
+	}
+
+	/**
+	 * @param  string $url
+	 * @param  mixed  $body
+	 *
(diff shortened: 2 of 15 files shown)


includes/chatbot/components/buttons/class-omise-chatbot-component-button-orderstatus.php
```diff
@@ -0,0 +1,13 @@
+<?php
+defined( 'ABSPATH' ) or die( 'No direct script access allowed.' );
+
+if ( class_exists( 'Omise_Chatbot_Component_Button_Orderstatus' ) ) {
+	return;
+}
+
+class Omise_Chatbot_Component_Button_Orderstatus extends Omise_Chatbot_Component_Button_Postback {
```
igroomgrim: If we create button class and extend from button_type(postback) class, then if we have a new button we will create new file right ?
Can we just create a button_type class ex. Omise_Chatbot_Component_Button_Postback and pass data to constructor instead create a new file, like
`$featured_product_button = new Omise_Chatbot_Component_Button_Postback("title", "payload")`
What do you think ?

guzzilar: @igroomgrim Sure, I think we should provide that too. (I mean, have both, specific classes for some often-used buttons that we can just easily reuse them and also have a way to create a custom button like you mentioned 👍? )

rule is: 
- if we use the same button more than 1 place, = have its own specific class.
- if we use only one time, just go with 
```
$featured_product_button = new Omise_Chatbot_Component_Button_Postback("title", "payload")
```

is that ok? or you think better to just have only `Omise_Chatbot_Component_Button_Postback` class without those `*_Orderstatus`, `*_Featuredproducts`, `*_Productcategory` prebuilt classes?

igroomgrim: I just see these class is look similar attribute and constructor (*_Orderstatus, *_Featuredproducts, *_Productcategory).
Then if we just only have Omise_Chatbot_Component_Button_Postback class it will be ok na.
🤖

guzzilar: @igroomgrim i c. I have these 3 classes (`Omise_Chatbot_Component_Button_*`) to avoid below use case.
screen shot 2560-09-25 at 11 17 55 pm

So we can refactor to
screen shot 2560-09-25 at 11 24 14 pm

That we can reduce mistake from human-error (like, change `label` / `payload` at 1 place, but forgot to update another places).

Anyway, so far I saw these buttons have been used just only 2 places. So I think it's ok to go with only `Omise_Chatbot_Component_Button_Postback` for now unless we have more place that need to reuse those buttons.

👆 so, 
```php
$this->components['template_button']
	->set_text( 'I\'m glad you visited How may I assist ?' )
	->add_buttons(
		array(
			new Omise_Chatbot_Component_Button_Postback( __( 'Featured products', 'omise' ), Omise_Chatbot_Payloads::ACTION_FEATURED_PRODUCTS ),
			new Omise_Chatbot_Component_Button_Postback( __( 'Product category', 'omise' ), Omise_Chatbot_Payloads::ACTION_PRODUCT_CATEGORY ),
			new Omise_Chatbot_Component_Button_Postback( __( 'Check order status', 'omise' ), Omise_Chatbot_Payloads::ACTION_ORDER_STATUS ),
		)
	);
```
👌  ?

guzzilar: btw, I still cheer to have its own class. 😂 
Because those `Omise_Chatbot_Facebook_Webhook_Event_Handler` classes don't need to taking care on `button title` or `action payload` at all, just initiate the right class and let's those button classes take care its own `title` and `payload`.

igroomgrim: Ohh after i see this 
screen shot 2560-09-25 at 11 24 14 pm
it look clean and save my eye 🤣 .
Then it ok if we separate to more classssss. 
:bowtie:

guzzilar: lol so no change now? 😁 

btw, ideally 👇 (if WordPress will go with `namespace` approach someday in the future)
screen shot 2560-09-25 at 11 41 19 pm

igroomgrim: `someday in the future` :tableflip:
It look very cool and cleannn. T.T

guzzilar:
👍

PR #94, opened 2019-01-17
[WIP] Refactoring plugin-initial code structure

guzzilar:
### 1. Objective

Omise-WooCommerce has always been coupled with WooCommerce plugin, which the plugin cannot be used as a stand-alone plugin and always require WooCommerce plugin to be activated first. 
However, the legacy-codebase itself wasn't really aware of a case where WooCommerce plugin is disabled as much as it should be.

Although we have prevented such cases #78, and #83. But by the current design of the plugin-initial step, it still remains some issue (#90) that could be ceased out by redesigning the initial-step with a proper awareness of WooCommerce plugin.

**Related issue(s)**:
- #90 Omise-WooCommerce 3.2 can't be activated.

#### 2. Description of change

[in progress]

#### 3. Quality assurance

**🔧 Environments:**

- **WordPress**: v5.0.3
- **WooCommerce**: v3.5.3
- **PHP version**: v5.6.30

**✏️ Details:**

The main test-case will be separated into 3 cases as follows:
1. Attempting to install Omise-WooCommerce without WooCommerce should give you a warning message that **Omise-WooCommerce requires WooCommerce plugin to be activated first.**
**Omise-WooCommerce** plugin can be activated, but all functions should be disabled, including no Omise menu at the WordPress admin-sidebar.

2. Install Omise-WooCommere plugin while WooCommerce plugin is enabled should give you a full-access to all Omise-WooCommerce feature, including the setting page.

3. Attempting to disable WooCommerce plugin while Omise-WooCommerce is still enabled should give you a same result as the case [1].

#### 4. Impact of the change

No

#### 5. Priority of change

Normal

#### 6. Additional Notes

No

Issue #90, opened 2018-07-22
Omise-WooCommerce 3.2 can't be activated.


jacekstan:
Environment:

WordPress 4.9.4
WooCommerce 3.3.4: active
Omise-WooCommerce 3.2
Result: Plugin **deactivated**. The Omise WooCommerce plugin requires **WooCommerce** to be installed and active.

Environment:

WordPress 4.9.4
WooCommerce 3.3.5: active
Omise-WooCommerce 3.2
Result: Plugin **deactivated**. The Omise WooCommerce plugin requires **WooCommerce** to be installed and active.

Environment:

WordPress 4.9.4
WooCommerce 3.3.4: active
Omise-WooCommerce 3.1
Result: activated

jacekstan:
that issue is happening for users who manually download `woocommerce` from github instead of using wordpress plugin installator.

this issue can be fixed by renaming woocommerce plugin directory from `woocommerce-X.X.X` to `woocommerce`

guzzilar:
@jacekstan 👍 Thanks for the diagnosing the issue. Shall I add tag `bug` here?

Issue #78, opened 2017-10-27
Fatal error, if install Omise plugin before WooCommerce


nimid:
WordPress can not be operated both frontend and backend, if Omise plugin has been installed and activated before installing WooCommerce.

The error message is shown in the screenshots below.

**Frontend**

<img width="1280" alt="frontend-error" src="https://user-images.githubusercontent.com/4145121/32128182-fb16c880-bba5-11e7-8ed4-8826b2a65d81.png">

**Backend**

<img width="1280" alt="backend-error" src="https://user-images.githubusercontent.com/4145121/32128188-040a99a8-bba6-11e7-81e8-1cfd7439f553.png">

WordPress 4.8.2
Omise WooCommerce 3.1

**Steps to reproduce**

- Fresh install WordPress
- Go to backend, Plugins > Add New
- Install Omise plugin (WooCommerce has not been installed)
- Activate Omise plugin

Although, WooCommerce has been installed and activated before Omise plugin, if WooCommerce has been deactivated, the error result is the same.

guzzilar:
@nimid Thanks for the report. I'll make patch to solve this on the next release.

guzzilar:
Note: this issue has been solved at release [v3.2](https://github.com/omise/omise-woocommerce/releases/tag/v3.2) (PR #83, #88).

PR #96, opened 2019-01-18
Refactoring plugin-initial code structure - part 3: Organizing Omise_Admin class

guzzilar:
> :warning: Important: This pull request requires **#95: "Refactoring plugin-initial code structure, part 2"** to be merged first.
#### 1. Objective

Omise-WooCommerce has always been coupled with WooCommerce plugin, which the plugin cannot be used as a stand-alone plugin and always require WooCommerce plugin to be activated first. 
However, the legacy-codebase itself wasn't really aware of a case where WooCommerce plugin is disabled as much as it should be.

Although we have prevented such cases #78, and #83. But by the current design of the plugin-initial step, it still remains some issue (#90) that could be ceased out by redesigning the initial-step with a proper awareness of WooCommerce plugin.

**Related issue(s)**:
- #90 Omise-WooCommerce 3.2 can't be activated.

#### Part 1

See, **#94: "Refactoring plugin-initial code structure, part 1"**

#### Part 2

See, **#95: "Refactoring plugin-initial code structure, part 2"**

#### Part 3

The last piece of code,
This part is aiming to organizing `Omise_Admin` class, by moving all admin-related code from `Omise` class back to `Omise_Admin`.

### 3. Quality assurance

**🔧 Environments:**

- **WordPress**: v5.0.3
- **WooCommerce**: v3.5.3
- **PHP version**: v5.6.30

**✏️ Details:**

The main test-case will be separated into 3 cases as follows:
1. Attempting to install Omise-WooCommerce without WooCommerce should give you a warning message that **Omise-WooCommerce requires WooCommerce plugin to be activated first.**
**Omise-WooCommerce** plugin can be activated, but all functions should be disabled, including no Omise menu at the WordPress admin-sidebar.
screen shot 2562-01-18 at 00 12 26

2. Install Omise-WooCommere plugin while WooCommerce plugin is enabled should give you a full-access to all Omise-WooCommerce feature, including the setting page.
screen shot 2562-01-17 at 23 32 33

3. Attempting to disable WooCommerce plugin while Omise-WooCommerce is still enabled should give you a same result as the case [1].
screen shot 2562-01-18 at 00 12 26

4. After the change, charge must be created, captured, refunded as usual.

### 4. Impact of the change

No

### 5. Priority of change

Normal

### 6. Additional Notes

No

### DIFF
--- a/omise-woocommerce.php
+++ b/omise-woocommerce.php
@@ -135,11 +135,8 @@ class Omise {
 	 */
 	protected function init_admin() {
 		if ( is_admin() ) {
-			require_once OMISE_WOOCOMMERCE_PLUGIN_PATH . '/includes/admin/class-omise-page-settings.php';
 			require_once OMISE_WOOCOMMERCE_PLUGIN_PATH . '/includes/class-omise-admin.php';
-
-			Omise_Admin::get_instance()->register_admin_menu();
-			add_filter( 'woocommerce_order_actions', array( $this, 'register_order_actions' ) );
+			Omise_Admin::get_instance()->init();
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


omise-woocommerce.php
```diff
@@ -29,48 +29,90 @@ class Omise {
...
 
 	/**
-	 * @since  3.0
+	 * Callback to display message about activation error
+	 *
+	 * @since  3.2
+	 */
+	public function init_error_messages(){
+		?>
+		<div class="error">
```
jonrandy: Again, this REALLY feels wrong. Does WP advocate this?

omise-woocommerce.php
```diff
@@ -29,48 +29,90 @@ class Omise {
...
+			<p><?php echo __( 'Omise WooCommerce plugin requires <strong>WooCommerce</strong> to be activated.', 'omise' ); ?></p>
```
jonrandy: Need to add 'The'

omise-woocommerce.php
guzzilar: Welcome to WordPress world.
I cannot tell if WP advocates for this but the WP codebase is even worse..
But yup, this is kind of common thing in WordPress (even WooCommerce and many plugins also go with this way everywhere in their code).

We can go a little advance by moving this thing into a template file and load it but I guess that would be out of scope for this pull request. Maybe we can discuss further in a future PR :)

omise-woocommerce.php
guzzilar: 👍 This will be automatically solved after merge #94

Issue #100, opened 2019-02-20
charge.complete event with status "failed" should not update completed orders


Thammada:
When a customer tries to pay using internet banking and does not complete the process, but created a new charge (using any method) successfully.
The order will be marked as completed when the second charge is successful, but when the first charge is timed out. The event `charge.complete` will be fired and the order will be marked as failed

The `charge.complete` event handler should check whether the order is already completed before updating the order as failed

PR #103, opened 2019-03-07
fixed internetbanking status fail after successful order

PlearnG:
#### 1. Objective

When a customer tries to pay using internet banking and does not complete the process, but created a new charge (using any method) successfully.
The order will be marked as completed when the second charge is successful, but when the first charge is timed out. The successful order will be updated as failed.

**Related information**:
Related issue(s): #100

#### 2. Description of change

The `charge.complete` event handler checks whether the order is already completed before updating the order as failed.

#### 3. Quality assurance

Tested manually by selecting internet banking as the payment method.

**🔧 Environments:**

i.e.
- **Platform version**: WooCommerce 3.4.4.
- **Omise plugin version**: Omise-WooCommerce 3.2.
- **PHP version**: 7.0.14.

**✏️ Details:**

- Select Internet banking as payment method without proceeding to payment
- Go back to checkout page
- Pay with Debit/Credit Card
- Wait for the first charge to time out

#### 4. Impact of the change
none

#### 5. Priority of change

Normal

### DIFF
--- a/includes/events/class-omise-event-charge-complete.php
+++ b/includes/events/class-omise-event-charge-complete.php
@@ -70,9 +70,9 @@ class Omise_Event_Charge_Complete {
 						$data->failure_message . ' (code: ' . $data->failure_code . ')'
 					)
 				);
-				if( $order->get_status() !== 'successful'){
+				if (!in_array( $order->get_status(), array('processing', 'completed' ))) {
 					$order->update_status( 'failed' );
-				}		
+				}
 				break;
 
 			case 'successful':
--- a/includes/events/class-omise-event-charge-complete.php
+++ b/includes/events/class-omise-event-charge-complete.php
@@ -70,8 +70,9 @@ class Omise_Event_Charge_Complete {
 						$data->failure_message . ' (code: ' . $data->failure_code . ')'
 					)
 				);
-
-				$order->update_status( 'failed' );
+				if( $order->get_status() !== 'successful'){
+					$order->update_status( 'failed' );
+				}		
 				break;
 
 			case 'successful':


guzzilar:
@PlearnG Thanks for submitting the patch for fixing the issue!
Let me work on it a moment and will give you back response soon. 

So far, I'm still considering whether adding a condition as the solution you have provided here, or having a parameter to say `payment_resolved: no` and it will be turned to `yes` only when it receives `successful` or `failed` status. As for the current solution still has one scenario remain where customer:
1. Select Internet banking as payment method without proceeding to payment.
2. Go back to checkout page.
3. Pay with another payment method (bank transfer, for instance).
4. Admin updates the order status manually to something else that is not in the condition.
5. When the `charge.complete` event arrives with `timeout` failure, the order status could still be updated to `failed`.
 
However, appreciated for the contribution!
Cheers.

Thammada:
@guzzilar how do you suggest we should add the `payment_resolved` parameter?
would using the action hook  [`woocommerce_payment_complete`](https://docs.woocommerce.com/wc-apidocs/source-class-WC_Order.html#125) be enough, which I think would apply only for `processing` and `completed` statuses (does `failed` actually need to be flagged as resolved?)

So on further inspection, the status `successful` that @PlearnG is checking against is actually incorrect, it should be `completed`

guzzilar:
@PlearnG @Thammada Hi, I just released a new Omise-WooCommerce v3.7 that contains a fix that related to this issue (https://github.com/omise/omise-woocommerce/pull/131).

I'm going to close this pull request at the moment, 
However, thanks for contributing 👍

PR #101, opened 2019-02-20
Refactoring, simplifying payment processors

guzzilar:
> Note: This pull request requires PR #111 to be merged first.

### 1. Objective

Alipay, Credit Card, Internet Banking payments are currently behaving almost the same in many aspects (except for some part such as charge-creating and handling the charge response).

However, they are completely separated from each other while some part of code could have been shared among each payment method to reduce those redundant code and make the payment processor more easy to maintain.

### 2. Description of change

Previously, each of payment method's class will have its own `process_payment` which have the same structure that can be written as follows:
```php
public function process_payment( $order_id ) {
    // 1. Loading WooCommerce Order from the $order_id into $order

    try {
        // 2. Logging an initiating message to WooCommerce Order
        // 3. Gathering all data and create a new charge
        // 4. Logging that charge has been created, attaching OmiseCharge.id as WooCommerce Order's transaction id.
        // 5. Validating a charge result, result will be depending on its charge.status (successful, failed, pending).
    } catch (Exception $e) {
        // 6. Notify an error message to the checkout page (wc_add_notice())
        // 7. Then, logging an error message to WooCommerce Order.
    }
}
```

By the above structure, we can simply take those redundant code out and group it into one method.
```php
class Omise_Payment {
    public function process_payment( $order_id ) {
        // 1. Loading WooCommerce Order from the $order_id into $order

        try {
            // 2. Logging an initiating message to WooCommerce Order

            $charge = $this->charge( $order_id, $order );

            // 4. Logging that charge has been created, attaching OmiseCharge.id as WooCommerce Order's transaction id.

            return $this->result( $order_id, $order, $charge );
        } catch (Exception $e) {
            // 6. Notify an error message to the checkout page (wc_add_notice())
            // 7. Then, logging an error message to WooCommerce Order.
        }
    }
}
```

And let those payment processors concern only _how to create its own charge_ and _validating a charge result._

```php
class Omise_Payment_Alipay extends Omise_Payment {
    public function charge( $order_id, $order ) { }
    public function result( $order_id, $order, $charge ) { }
}

class Omise_Payment_Creditcard extends Omise_Payment {
    public function charge( $order_id, $order ) { }
    public function result( $order_id, $order, $charge ) { }
}

class Omise_Payment_Internetbanking extends Omise_Payment {
    public function charge( $order_id, $order ) { }
    public function result( $order_id, $order, $charge ) { }
}
```

[...]

PR #108, opened 2019-03-28
Refactoring, unify Omise key(s)-defining into one place.

guzzilar:
#### 1. Objective

Currently in Omise-WooCommerce code base, the public key and the secret key have been spread all over the places. By that, it has a high risk and a potential of any bug from code maintenance in the future (as there are many places that need to be taken care of).  

This pull request is to unifying Omise-Keys declaration into one place to reduce any potential bug & code mess.

#### 2. Description of change

- Defining Omise secret key and public key at the plugin initial step.
- Removing all related code.

#### 3. Quality assurance

**🔧 Environments:**

- **WordPress**: v5.1.1.
- **WooCommerce**: v3.5.7.
- **PHP version**: v7.3.3.

**✏️ Details:**

All of cases that we should test would be really huge as this pull request is removing all key-assigning out from all of Omise class. However, the main test could be listed as below:

1. Make sure that Alipay payment method works properly by creating a new charge with Alipay payment method.

2. Make sure that Internet Banking payment method works properly by creating a new charge with Internet Banking payment method.

3. Make sure that Credit Card payment method works properly by creating a new charge with Credit Card payment method on both `auth only` and `auto capture`.

4. Make sure that an order can be funded properly

5. Make sure that user can save a card properly.

6. Make sure that user can delete a card properly.

#### 4. Impact of the change

No

#### 5. Priority of change

Normal

#### 6. Additional Notes

This change will help on https://github.com/omise/omise-woocommerce/pull/101 as well (as there is no keys tied up in those methods anymore after this change).

### DIFF
--- a/includes/class-omise-setting.php
+++ b/includes/class-omise-setting.php
@@ -8,9 +8,35 @@ if ( class_exists( 'Omise_Setting' ) ) {
 
 class Omise_Setting {
 	/**
+	 * The Omise_Setting Instance.
+	 *
+	 * @since 3.4
+	 *
+	 * @var   \Omise_Setting
+	 */
+	protected static $the_instance = null;
+
+	/**
 	 * @var null | array
 	 */
 	public $settings;
+
+	/**
+	 * The Omise_Setting Instance.
+	 *
+	 * @since  3.4
+	 *
+	 * @static
+	 *
+	 * @return \Omise_Setting - The instance.
+	 */
+	public static function instance() {
+		if ( is_null( self::$the_instance ) ) {
+			self::$the_instance = new self();
+		}
+
+		return self::$the_instance;
+	}
 
 	/**
 	 * @since 3.1
--- a/includes/class-omise-wc-myaccount.php
+++ b/includes/class-omise-wc-myaccount.php
@@ -4,7 +4,7 @@ defined( 'ABSPATH' ) or die( 'No direct script access allowed.' );
 if ( ! class_exists( 'Omise_MyAccount' ) ) {
 	class Omise_MyAccount {
 		private static $instance;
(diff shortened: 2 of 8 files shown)


includes/class-omise-setting.php
```diff
@@ -7,11 +7,37 @@
 }
 
 class Omise_Setting {
```
jonrandy: Should really be called "Settings" - otherwise the name suggests a class that deals with one setting only

includes/class-omise-wc-myaccount.php
```diff
@@ -103,7 +81,7 @@ public function register_omise_my_account_scripts() {
 				'omise-myaccount-card-handler',
 				'omise_params',
 				array(
-					'key'             => $this->public_key,
+					'key'             => Omise()->setting->public_key(),
 					'ajax_url'        => admin_url( 'admin-ajax.php' ),
```
jonrandy: Not sure, but should this be `setting()`?

jonrandy: Again, this should definitely be named `settings`

guzzilar:
@jonrandy @jacstn updated, can you help check https://github.com/omise/omise-woocommerce/pull/108/commits/f1cc6db126c32e381a86d93f7fa8e1dab02182d2

However, renaming `Omise_Setting` class to `Omise_Settings` would touch more on an irrelevant part of code than its actual purpose. Agreed to change, but I think it's better to submit another pr for that only (I'll have a clean-up pr later on before the release v3.4).

guzzilar:
Screen Shot 2562-03-28 at 14 27 33

guzzilar:
Note: Updated Quality Assurance section. All tested.

PR #111, opened 2019-03-30
Removing all redundant code regarding to WC Order transaction ID assignment.

guzzilar:
#### 1. Objective

Omise-WooCommerce, back to the time, was seeking for a way to retrieve Omise charge id from WooCommerce Order that has been placed. So we've decided to create a new custom-field inside a WC order.

Now, it comes to realized that actually we could simply integrate with WC order's transaction id.
This pull request is aiming to remove the mess.

#### 2. Description of change

- No more attaching Omise charge id with a custom field inside Order item.
- Refactoring the code accordingly.

#### 3. Quality assurance

**🔧 Environments:**

- **WordPress**: v5.1.1
- **WooCommerce**: v3.5.7
- **PHP version**: 7.3.3

**✏️ Details:**

Make sure that all 3 payment methods could assign Omise charge id to its order properly.
**Alipay**
Screen Shot 2562-03-30 at 13 42 21

**Credit Card**
Screen Shot 2562-03-30 at 13 44 18

**Internet Banking**
Screen Shot 2562-03-30 at 13 43 54

Also make sure that a feature that requires _retrieving charge id from an order item_ would function properly.
**Capture a charge**
Screen Shot 2562-03-30 at 13 45 23

#### 4. Impact of the change

Not in particularly, but 3rd-party plugins may want to have a look if anyone has implemented with the charge-id custom field that this pull request is removing (not affect, just, deprecated code and would be nice to update the codebase)

#### 5. Priority of change

Normal

#### 6. Additional Notes

The reason that we cannot completely get rid of legacy code is because there might be someone or some 3rd-party plugins that have implemented with those code already.

Would like to keep it for a while or until WooCommerce has a new major release.

### DIFF
--- a/includes/gateway/class-omise-payment-creditcard.php
+++ b/includes/gateway/class-omise-payment-creditcard.php
@@ -297,18 +297,10 @@ function register_omise_creditcard() {
 
 				$order->add_order_note( sprintf( __( 'Omise: Charge (ID: %s) has been created', 'omise' ), $charge['id'] ) );
 
-				$this->attach_charge_id_to_order( $charge['id'] );
+				$this->set_order_transaction_id( $charge['id'] );
 
 				if ( Omise_Charge::is_failed( $charge ) ) {
 					throw new Exception( Omise_Charge::get_error_message( $charge ) );
-				}
-
-				/** backward compatible with WooCommerce v2.x series **/
-				if ( version_compare( WC()->version, '3.0.0', '>=' ) ) {
-					$order->set_transaction_id( $charge['id'] );
-					$order->save();
-				} else {
-					update_post_meta( $order->id, '_transaction_id', $charge['id'] );
 				}
 
 				if ( $this->omise_3ds ) {
--- a/includes/gateway/class-omise-payment-alipay.php
+++ b/includes/gateway/class-omise-payment-alipay.php
@@ -106,17 +106,9 @@ function register_omise_alipay() {
 
 				switch ( $charge['status'] ) {
 					case 'pending':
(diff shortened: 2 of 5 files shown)


includes/gateway/class-omise-payment-creditcard.php
```diff
@@ -297,20 +297,12 @@ public function process_payment( $order_id ) {
 
 				$order->add_order_note( sprintf( __( 'Omise: Charge (ID: %s) has been created', 'omise' ), $charge['id'] ) );
 
-				$this->attach_charge_id_to_order( $charge['id'] );
+				$this->set_order_transaction_id( $charge['id'] );
```
jacstn: why this name was changed?

guzzilar: @jacstn  Originally it was named `attach_charge_id_to_order` as its intention, to attach Omise charge id to WC's order item so we can retrieve it back later for those offsite payments (3-D Secure included).

Then, it comes to the second version, where I integrated `Omise charge id` with `transaction id` feature provided by WC as you can see from the following code:
```php
if ( version_compare( WC()->version, '3.0.0', '>=' ) ) {
	$order->set_transaction_id( $charge['id'] );
	$order->save();
} else {
	update_post_meta( $order->id, '_transaction_id', $charge['id'] );
}
```
However, these 2 code are redundant to each other. So, I removed these 2 code, `attach_charge_id_to_order`, and the above code. Then having `set_order_transaction_id` method, that simply to the job as its name (set order transaction id).

PR #115, opened 2019-05-29
Installment, display instalment monthly amount & interest fee at the checkout page.

guzzilar:
#### 1. Objective

As @jonrandy and @jacstn have already mentioned at `omise/omise-magento` project:

> This PR essentially adds everything we can for now to ensure that should a customer go to the instalment provider's page, the payment options they have selected will be valid and the chances are they will proceed with the payment (hopefully drastically minimising the chances of the cancellation issue).

This pull request was inspired by them (thanks guys) 
_and that saved me some time to write my own objective_ 😂 

#### 2. Description of change

There are 2 big things going on in this pull request.
1. Introducing a brand-new `Backend` class as for a place to house a specific data and information that could be used to handle and manipulate all necessary data before passing it to front-end or before process a payment.

2. Displaying an installment monthly amount and an interest fee at the checkout page, Installment payment form.
  display-interest-1
  display-interest-2
  display-interest-3

#### 3. Quality assurance

**🔧 Environments:**

This pull request was developed and tested under these machine's environment:

- **WordPress**: v5.2.1
- **WooCommerce**: v3.6.3
- **PHP**: v7.3.3

**✏️ Details:**

You can either run PHPUnit to make sure that all of the calculations have done right

Screen Shot 2562-05-29 at 17 03 35

Or check its result (the same as the above 3 screenshots) by installing this pull request, enable Installment payment method, and go to the check out page.

#### 4. Impact of the change

No (besides user interface has changed).

#### 5. Priority of change

Normal

#### 6. Additional Notes

No

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,9 @@
+{
+    "name": "omise/omise-woocommerce",
+    "description": "An official payment extension which provides support for Omise payment gateway for store builders working on the WooCommerce platform.",
+    "homepage": "https://www.omise.co/",
+    "license": "MIT",
+    "require-dev": {
+        "phpunit/phpunit": "^8.1"
+    }
+}
--- a/includes/backends/class-omise-backend-installment.php
+++ b/includes/backends/class-omise-backend-installment.php
@@ -22,35 +22,35 @@ class Omise_Backend_Installment extends Omise_Backend {
 			'installment_first_choice' => array(
 				'bank_code'          => 'first_choice',
 				'title'              => __( 'Krungsri First Choice', 'omise' ),
-				'interest_rate'      => 0.013,
+				'interest_rate'      => 1.3,
 				'min_allowed_amount' => 300.00,
 			),
 
 			'installment_bay' => array(
 				'bank_code'          => 'bay',
 				'title'              => __( 'Krungsri', 'omise' ),
-				'interest_rate'      => 0.008,
+				'interest_rate'      => 0.8,
 				'min_allowed_amount' => 300.00,
 			),
 
 			'installment_ktc' => array(
 				'bank_code'          => 'ktc',
 				'title'              => __( 'Krungthai Card (KTC)', 'omise' ),
-				'interest_rate'      => 0.008,
+				'interest_rate'      => 0.8,
 				'min_allowed_amount' => 300.00,
 			),
 
 			'installment_bbl' => array(
 				'bank_code'          => 'bbl',
 				'title'              => __( 'Bangkok Bank', 'omise' ),
-				'interest_rate'      => 0.008,
+				'interest_rate'      => 0.8,
 				'min_allowed_amount' => 500.00,
 			),
 
 			'installment_kbank' => array(
 				'bank_code'          => 'kbank',
 				'title'              => __( 'Kasikorn Bank', 'omise' ),
-				'interest_rate'      => 0.0065,
+				'interest_rate'      => 0.65,
 				'min_allowed_amount' => 500.00,
 			),
 		);
@@ -72,7 +72,7 @@ class Omise_Backend_Installment extends Omise_Backend {
 
 			$provider->provider_code   = str_replace( 'installment_', '', $provider->_id );
 			$provider->provider_name   = isset( $provider_detail ) ? $provider_detail['title'] : strtoupper( $provider->code );
-			$provider->interest_rate   = $this->capabilities()->is_zero_interest() ? 0 : ( $provider_detail['interest_rate'] * 100 );
+			$provider->interest_rate   = $this->capabilities()->is_zero_interest() ? 0 : ( $provider_detail['interest_rate'] );
 			$provider->available_plans = $this->get_available_plans(
 				$purchase_amount,
 				$provider->allowed_installment_terms,
@@ -116,12 +116,12 @@ class Omise_Backend_Installment extends Omise_Backend {
 	/**
 	 * @param  float $purchase_amount
 	 * @param  int   $term_length      A length of a given installment term.
-	 * @param  float $interest_rate    its value can be '0' if merchant absorbs an interest.
+	 * @param  float $interest_rate    Its value can be '0' if merchant absorbs an interest.
 	 *
 	 * @return float  of a installment monthly payment (round up to 2 decimals).
 	 */
 	public function calculate_monthly_payment_amount( $purchase_amount, $term_length, $interest_rate ) {
-		$interest = $purchase_amount * $term_length * $interest_rate;
+		$interest = $purchase_amount * $interest_rate * $term_length / 100;
 		return round( ( $purchase_amount + $interest ) / $term_length, 2 );
 	}
 }
--- a/includes/backends/class-omise-backend-installment.php
+++ b/includes/backends/class-omise-backend-installment.php
@@ -1,0 +1,123 @@
+<?php
+/**
+ * Note: all of calculations in this class are based only on Thailand VAT and fee
+ *       as currently Installment feature supported only for merchants
+ *       that have registered with Omise Thailand account.
+ *
+ * @since 3.4
+ *
+ * @method public initiate
+ * @method public get_available_providers
+ * @method public filter_available_terms
+ * @method public calculate_monthly_payment_amount
+ */
+class Omise_Backend_Installment extends Omise_Backend {
+	/**
+	 * @var array  of known installment providers.
+	 */
+	protected static $providers = array();
+
+	public function initiate() {
+		self::$providers = array(
+			'installment_first_choice' => array(
+				'bank_code'          => 'first_choice',
+				'title'              => __( 'Krungsri First Choice', 'omise' ),
+				'interest_rate'      => 0.013,
+				'min_allowed_amount' => 300.00,
+			),
+
+			'installment_bay' => array(
+				'bank_code'          => 'bay',
+				'title'              => __( 'Krungsri', 'omise' ),
+				'interest_rate'      => 0.008,
+				'min_allowed_amount' => 300.00,
+			),
+
+			'installment_ktc' => array(
+				'bank_code'          => 'ktc',
+				'title'              => __( 'Krungthai Card (KTC)', 'omise' ),
+				'interest_rate'      => 0.008,
+				'min_allowed_amount' => 300.00,
+			),
+
+			'installment_bbl' => array(
+				'bank_code'          => 'bbl',
+				'title'              => __( 'Bangkok Bank', 'omise' ),
+				'interest_rate'      => 0.008,
+				'min_allowed_amount' => 500.00,
+			),
+
+			'installment_kbank' => array(
+				'bank_code'          => 'kbank',
+				'title'              => __( 'Kasikorn Bank', 'omise' ),
+				'interest_rate'      => 0.0065,
+				'min_allowed_amount' => 500.00,
+			),
+		);
+	}
+
+	/**
+	 * @param  string $currency
+	 * @param  float  $purchase_amount
+	 *
+	 * @return array  of an available installment providers
+	 */
+	public function get_available_providers( $currency, $purchase_amount ) {
+		// Note: as installment payment at the moment only supports for THB currency, so the 
+		//       $purchase_amount is multiplied with 100 to convert the amount into subunit (satang).
+		$providers = $this->capabilities()->getInstallmentBackends( $currency, ( $purchase_amount * 100 ) );
+
+		foreach ( $providers as &$provider ) {
+			$provider->provider_code             = str_replace( 'installment_', '', $provider->_id );
+			$provider->provider_name             = isset( self::$providers[ $provider->_id ] ) ? self::$providers[ $provider->_id ]['title'] : strtoupper( $provider->code );
+			$provider->allowed_installment_terms = $this->filter_available_terms( $provider->allowed_installment_terms, self::$providers[ $provider->_id ], $purchase_amount );
+		}
+
+		return $providers;
+	}
+
+	/**
+	 * @param  array $available_terms
+	 * @param  array $provider_detail
+	 * @param  float $purchase_amount
+	 *
+	 * @return array  of an filtered available terms
+	 */
+	public function filter_available_terms( $available_terms, $provider_detail, $purchase_amount ) {
+		$filtered_available_terms = array();
+
+		sort( $available_terms );
+
+		for ( $i = 0; $i < count( $available_terms ); $i++ ) {
+			$term           = $available_terms[ $i ];
+			$monthly_amount = $this->calculate_monthly_payment_amount(
+				$purchase_amount,
+				$term,
+				$this->capabilities()->is_zero_interest() ? 0 : $provider_detail['interest_rate']
+			);
+
+			if ( $monthly_amount < $provider_detail['min_allowed_amount'] ) {
+				break;
+			}
+
+			$filtered_available_terms[ $i ] = array(
+				'term'           => $term,
+				'monthly_amount' => $monthly_amount
+			);
+		}
+
+		return $filtered_available_terms;
+	}
+
+	/**
+	 * @param  float $purchase_amount
+	 * @param  int   $term             installment term.
+	 * @param  float $interest_rate    its value can be '0' if merchant absorbs an interest.
+	 *
+	 * @return float  of a installment monthly payment (round up to 2 decimals).
+	 */
+	public function calculate_monthly_payment_amount( $purchase_amount, $term, $interest_rate ) {
+		$interest = $purchase_amount * $term * $interest_rate;
+		return round( ( $purchase_amount + $interest ) / $term, 2 );
+	}
+}
--- a/includes/backends/class-omise-backend-installment.php
+++ b/includes/backends/class-omise-backend-installment.php
@@ -70,10 +70,10 @@ class Omise_Backend_Installment extends Omise_Backend {
 		foreach ( $providers as &$provider ) {
 			$provider_detail = self::$providers[ $provider->_id ];
 
-			$provider->provider_code      = str_replace( 'installment_', '', $provider->_id );
-			$provider->provider_name      = isset( $provider_detail ) ? $provider_detail['title'] : strtoupper( $provider->code );
-			$provider->interest_rate      = $this->capabilities()->is_zero_interest() ? 0 : ( $provider_detail['interest_rate'] * 100 );
-			$provider->available_plans    = $this->get_available_plans(
+			$provider->provider_code   = str_replace( 'installment_', '', $provider->_id );
+			$provider->provider_name   = isset( $provider_detail ) ? $provider_detail['title'] : strtoupper( $provider->code );
+			$provider->interest_rate   = $this->capabilities()->is_zero_interest() ? 0 : ( $provider_detail['interest_rate'] * 100 );
+			$provider->available_plans = $this->get_available_plans(
 				$purchase_amount,
 				$provider->allowed_installment_terms,
 				$provider->interest_rate,
--- a/includes/backends/class-omise-backend-installment.php
+++ b/includes/backends/class-omise-backend-installment.php
@@ -85,12 +85,11 @@ class Omise_Backend_Installment extends Omise_Backend {
 	 * @return array  of an filtered available terms
 	 */
 	public function get_valid_terms( $available_terms, $provider_detail, $purchase_amount ) {
-		$filtered_available_terms = array();
+		$valid_terms = array();
 
 		sort( $available_terms );
 
-		foreach ( $available_terms as $key => $available_term ) {
-			$term           = $available_terms[ $key ];
+		foreach ( $available_terms as $term ) {
 			$monthly_amount = $this->calculate_monthly_payment_amount(
 				$purchase_amount,
 				$term,
@@ -101,13 +100,13 @@ class Omise_Backend_Installment extends Omise_Backend {
 				break;
 			}
 
-			$filtered_available_terms[ $key ] = array(
+			$valid_terms[] = array(
 				'term'           => $term,
 				'monthly_amount' => $monthly_amount
 			);
 		}
 
-		return $filtered_available_terms;
+		return $valid_terms;
 	}
 
 	/**
--- a/includes/backends/class-omise-backend-installment.php
+++ b/includes/backends/class-omise-backend-installment.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Note: all of calculations in this class are based only on Thailand VAT and fee
+ * Note: all calculations in this class are based only on Thailand VAT and fee
  *       as currently Installment feature supported only for merchants
  *       that have registered with Omise Thailand account.
  *
@@ -8,7 +8,7 @@
  *
  * @method public initiate
  * @method public get_available_providers
- * @method public get_valid_terms
+ * @method public get_available_plans
  * @method public calculate_monthly_payment_amount
  */
 class Omise_Backend_Installment extends Omise_Backend {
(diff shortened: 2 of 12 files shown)


includes/backends/class-omise-backend-installment.php
```diff
@@ -0,0 +1,124 @@
...
+ * @method public calculate_monthly_payment_amount
+ */
+class Omise_Backend_Installment extends Omise_Backend {
+	/**
+	 * @var array  of known installment providers.
+	 */
+	protected static $providers = array();
+
+	public function initiate() {
+		self::$providers = array(
```
guzzilar: I intended to have this `initiate` here as normally WordPress project's behaviour is to have a place where 3rd-party developers can hook and alter its value (something like `apply_filter()`). – _just leave a room for its possibility but don't have any real request or a use-case at the moment._

includes/backends/class-omise-backend-installment.php
```diff
@@ -0,0 +1,124 @@
...
+	 * @param  float $purchase_amount
+	 *
+	 * @return array  of an filtered available terms
+	 */
+	public function filter_available_terms( $available_terms, $provider_detail, $purchase_amount ) {
+		$filtered_available_terms = array();
+
+		sort( $available_terms );
+
+		for ( $i = 0; $i < count( $available_terms ); $i++ ) {
```
jonrandy: Would `forEach` be simpler?

includes/backends/class-omise-backend-installment.php
```diff
@@ -0,0 +1,124 @@
...
+	 * @param  array $available_terms
+	 * @param  array $provider_detail
+	 * @param  float $purchase_amount
+	 *
+	 * @return array  of an filtered available terms
+	 */
+	public function filter_available_terms( $available_terms, $provider_detail, $purchase_amount ) {
+		$filtered_available_terms = array();
+
+		sort( $available_terms );
```
jonrandy: Should a filter function be sorting?

includes/backends/class-omise-backend-installment.php
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	/**
+	 * @param  array $available_terms
+	 * @param  array $provider_detail
+	 * @param  float $purchase_amount
+	 *
+	 * @return array  of an filtered available terms
+	 */
+	public function filter_available_terms( $available_terms, $provider_detail, $purchase_amount ) {
```
jonrandy: Why not `array_filter($available_terms, $valid_term_callable)`? That would be the normal way to filter an array?

includes/backends/class-omise-backend-installment.php
```diff
@@ -0,0 +1,124 @@
...
+				break;
+			}
+
+			$filtered_available_terms[ $i ] = array(
+				'term'           => $term,
+				'monthly_amount' => $monthly_amount
+			);
+		}
+
+		return $filtered_available_terms;
```
jonrandy: After reading the whole function it seems really that it is possibly poorly named - as it is transforming the array to have different content, as well as sorting and filtering.

tests/unit/includes/backends/class-omise-backend-installment-test.php
```diff
@@ -0,0 +1,151 @@
...
+class Omise_Backend_Installment_Test extends TestCase {
+	public static function setUpBeforeClass(): void {
+		Omise_Unit_Test::include_class( 'backends/class-omise-backend.php' );
+		Omise_Unit_Test::include_class( 'backends/class-omise-backend-installment.php' );
+	}
+
+	/**
+	 * @test
+	 */
+	public function filtering_bbl_installment_term() {
```
jonrandy: Might want to rename this too based on previous comments

tests/unit/includes/backends/class-omise-backend-installment-test.php
```diff
@@ -0,0 +1,151 @@
...
+ * @see includes/class-omise-capabilities
+ */
+class Omise_Capabilities {
+	/**
+	 * @var self
+	 */
+	protected static $the_instance = null;
+
+	public static function retrieve() {
+		self::$the_instance = new self();
```
jonrandy: Should this be `self::$the_instance = self::$the_instance ?: new self();`?

includes/backends/class-omise-backend-installment.php
guzzilar: @jonrandy do you have any suggestion name? Maybe I use `filter` in a wrong meaning here. 
The intention of the code is that you throw the whole list of all possible terms in, then get only available terms that matched with the requirement out.

tests/unit/includes/backends/class-omise-backend-installment-test.php
jacstn: tested a branch,  it works well

jacstn:
tested a branch, it works well

templates/payment/form-installment.php
```diff
@@ -10,15 +10,23 @@
...
 							<select id="<?php echo $backend->_id; ?>_installment_terms" name="<?php echo $backend->_id; ?>_installment_terms" class="installment-term-select-box">
 								<option>Select term</option>
-								<?php foreach ( $backend->allowed_installment_terms as $term ) : ?>
-									<option value="<?php echo $term; ?>"><?php echo $term; ?> <?php echo __('months'); ?></option>
+								<?php foreach ( $backend->allowed_installment_terms as $installment_term ) : ?>
+									<option value="<?php echo $installment_term['term']; ?>"><?php echo $installment_term['term']; ?> <?php echo __('months'); ?> ( <?php echo wc_price( $installment_term['monthly_amount'] ); ?> / <?php echo __('months'); ?> )</option>
 								<?php endforeach; ?>
 							</select>
+							<?php if ( ! $viewData['is_zero_interest'] ): ?>
+								<br/><span class="omise-installment-interest-rate">( <?php echo __( 'interest', 'omise' ); ?> <?php echo $backend->interest; ?>% )</span>
```
jacstn: you never know how it will be exactly translated to other languages.
which word will be where depends on the language.

generating strings to translate should be made using mechanisms from wordpress
https://codex.wordpress.org/I18n_for_WordPress_Developers

example:
```
<p>
<?php
$url = 'http://example.com';
$link = sprintf( wp_kses( __( 'Check out this link to my <a href="%s">website</a> made with WordPress.', 'my-text-domain' ), array(  'a' => array( 'href' => array() ) ) ), esc_url( $url ) );
echo $link;
?>
</p>
```

includes/backends/class-omise-backend-installment.php
jonrandy: Well, the `type` of what is in the array coming out here is different to what is going in - so it is doing more than just filtering. It's also sorting. Maybe something like 'get_valid_term_options` would be better, where we consider a `term_option` to be the array containing the number of `terms` (plural important) and the `monthly_amount`.

Alternatively if you want to go all functional, you could have one function to transform a `term` to a `term_option`, and another to check if a term option was valid... then, in pseudocode:

```
availableTerms = terms.sort().map(convertToTermOptions).filter(validTermOption)
```

Although this isn't as efficient as your code above correctly stops adding any more valid term options as soon as it hits the first one that is below minimum

templates/payment/form-installment.php
jonrandy: @jacstn Well spotted. Eagle eyes 👀

composer.json
```diff
@@ -0,0 +1,9 @@
+{
+    "name": "omise/omise-woocommerce",
+    "description": "An official payment extension which provides support for Omise payment gateway for store builders working on the WooCommerce platform.",
```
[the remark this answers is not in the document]
guzzilar: @jacstn I copied this one from the first paragraph of the [README.md](https://github.com/omise/omise-woocommerce/blob/master/README.md) file.
Maybe to change this one, maybe we should have another PR to update all related places?

includes/backends/class-omise-backend-installment.php
guzzilar: @jonrandy  Somehow forgot that `foreach` exists 😂 
Thanks.

templates/payment/form-installment.php
guzzilar: @jacstn  👍 nice catch, thanks.
I'm changing it to
```php
echo sprintf(  __( '( interest %g%% )', 'omise' ), $backend->interest );
```

includes/backends/class-omise-backend-installment.php
guzzilar: @jonrandy Thanks for the explanation 👍 I see your point why `filter` doesn't make sense now.
Just wonder would it be any confusion between `allowed`, `available`, and `valid`?

jonrandy: I have no problem with any of those terms, it wasn't a conscious decision to change to to `valid`

composer.json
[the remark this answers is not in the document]
guzzilar: @jacstn Hi, thanks for approving the PR. I'm going to resolve this conversation, you ok?

includes/backends/class-omise-backend-installment.php
guzzilar: @jonrandy Hi, I didn't change this one at the recent commit as we were discussing about this misled-method name.
Let me know if you still have any further opinion about sorting here. : D

composer.json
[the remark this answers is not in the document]
guzzilar: Got a confirmation from @jacstn, resolved 😄

guzzilar:
FYI, @jacstn Hi, I've made a little change after you approved.
- Omise_Backend_Installment::get_valid_terms(), leaning the method (more efficient): https://github.com/omise/omise-woocommerce/pull/115/commits/810520a7bc33adbaf2b17fbc98c065da75fc9fdd

Feel free to review if you want :D 
Thanks

includes/backends/class-omise-backend-installment.php
jonrandy: I'm still not overly keen on `term` in the code referring to 2 different things - both a single number representing the number of months (which isn't what term means - a better name would be `terms` or `term_count`), and an associative array containing the number of terms and the monthly amount

includes/backends/class-omise-backend-installment.php
```diff
@@ -0,0 +1,123 @@
...
+				'monthly_amount' => $monthly_amount
+			);
+		}
+
+		return $valid_terms;
+	}
+
+	/**
+	 * @param  float $purchase_amount
+	 * @param  int   $term             installment term.
```
jonrandy: Should be `terms`

includes/class-omise-capabilities.php
```diff
@@ -45,4 +45,11 @@ public function getInstallmentBackends( $currency = '', $amount = null ) {
 
 		return $this->capabilities->getBackends( $params );
 	}
+
+	/**
+	 * @return bool  True if merchant absorbs an interest or else, false.
```
jonrandy: 'the interest'

includes/backends/class-omise-backend-installment.php
guzzilar: @jonrandy you mean something like these 2 cases 👇 ?

" This BBL has 3 installment terms you can choose, which are '3', '6', '9'. "

and

available_terms = "3", "6", "9"
term 3, term 6, term 9.

jonrandy: The available terms are a term of 3 months, a term of 6 months, and a term of 9 months. 'Term' is the time period itself, not it's length.

Aside from the naming issue, my problem with the above function was the use of the word term to refer to both:

a) The length of the term over which the instalments are paid

and 

b) A construct representing the details of a particular instalment plan (the monthly payment amount and the term length)

This dual use of the same word seems wrong

guzzilar: Uhmm... my English is error today, any suggestion? @jonrandy 😂 
Where it should be replaced with `term_length` or `term_months`?

guzzilar:
```php
$provider->allowed_installment_terms = $this->get_valid_terms(...);

public function get_valid_terms() {
    $valid_terms = array();

    return $valid_terms;
}
```
👆 These terms context seem ok to me. Maybe inside `foreach` one?

jonrandy: A better name may even be `get_valid_instalment_plans`, given what is actually being returned

jonrandy: An attempt at a rewrite... might not be 100% perfect.

```php
<?php
/**
 * Note: all calculations in this class are based only on Thailand VAT and fee
 *       as currently Installment feature supported only for merchants
 *       that have registered with Omise Thailand account.
 *
 * @since 3.4
 *
 * @method public initiate
 * @method public get_available_providers
 * @method public get_valid_plans
 * @method public calculate_monthly_payment_amount
 */
class Omise_Backend_Installment extends Omise_Backend {
	/**
	 * @var array  of known installment providers.
	 */
	protected static $providers = array();

	public function initiate() {
		self::$providers = array(
			'installment_first_choice' => array(
				'bank_code'          => 'first_choice',
				'title'              => __( 'Krungsri First Choice', 'omise' ),
				'interest_rate'      => 0.013,
				'min_allowed_amount' => 300.00,
			),

			'installment_bay' => array(
				'bank_code'          => 'bay',
				'title'              => __( 'Krungsri', 'omise' ),
				'interest_rate'      => 0.008,
				'min_allowed_amount' => 300.00,
			),

			'installment_ktc' => array(
				'bank_code'          => 'ktc',
				'title'              => __( 'Krungthai Card (KTC)', 'omise' ),
				'interest_rate'      => 0.008,
				'min_allowed_amount' => 300.00,
			),

			'installment_bbl' => array(
				'bank_code'          => 'bbl',
				'title'              => __( 'Bangkok Bank', 'omise' ),
				'interest_rate'      => 0.008,
				'min_allowed_amount' => 500.00,
			),

			'installment_kbank' => array(
				'bank_code'          => 'kbank',
				'title'              => __( 'Kasikorn Bank', 'omise' ),
				'interest_rate'      => 0.0065,
				'min_allowed_amount' => 500.00,
			),
		);
	}

	/**
	 * @param  string $currency
	 * @param  float  $purchase_amount
	 *
	 * @return array  of an available installment providers
	 */
	public function get_available_providers( $currency, $purchase_amount ) {
		// Note: as installment payment at the moment only supports for THB currency, so the 
		//       $purchase_amount is multiplied with 100 to convert the amount into subunit (satang).
		$providers = $this->capabilities()->getInstallmentBackends( $currency, ( $purchase_amount * 100 ) );

		foreach ( $providers as &$provider ) {
			$provider->provider_code             = str_replace( 'installment_', '', $provider->_id );
			$provider->provider_name             = isset( self::$providers[ $provider->_id ] ) ? self::$providers[ $provider->_id ]['title'] : strtoupper( $provider->code );
			$provider->allowed_installment_plans = $this->get_valid_plans( $provider->allowed_installment_terms, self::$providers[ $provider->_id ], $purchase_amount );
			$provider->interest_rate             = $this->capabilities()->is_zero_interest() ? 0 : ( self::$providers[ $provider->_id ]['interest_rate'] * 100 );
		}

		return $providers;
	}

	/**
	 * @param  array $available_term_lengths (months)
	 * @param  array $provider_detail
	 * @param  float $purchase_amount
	 *
	 * @return array  of available installment plans
	 */
	public function get_valid_plans( $available_term_lengths, $provider_detail, $purchase_amount ) {
		$valid_plans = array();

		sort( $available_term_lengths );

		foreach ( $available_term_lengths as $term_length ) {
			$monthly_amount = $this->calculate_monthly_payment_amount(
				$purchase_amount,
				$term_length,
				// *** Wasn't this is_zero_interest check done before? ***
				$this->capabilities()->is_zero_interest() ? 0 : $provider_detail['interest_rate']
			);

			if ( $monthly_amount < $provider_detail['min_allowed_amount'] ) {
				break;
			}

			$valid_plans[] = array(
				'term_length'           => $term_length,
				'monthly_amount' 				=> $monthly_amount
			);
		}

		return $valid_plans;
	}

	/**
	 * @param  float $purchase_amount
	 * @param  int   $term             installment term.
	 * @param  float $interest_rate    its value can be '0' if merchant absorbs an interest.
	 *
	 * @return float  of a installment monthly payment (round up to 2 decimals).
	 */
	public function calculate_monthly_payment_amount( $purchase_amount, $term, $interest_rate ) {
		$interest = $purchase_amount * $term_length * $interest_rate;
		return round( ( $purchase_amount + $interest ) / $term_length, 2 );
	}
}
```

includes/backends/class-omise-backend-installment.php
```diff
@@ -0,0 +1,123 @@
...
+	public function get_valid_terms( $available_terms, $provider_detail, $purchase_amount ) {
+		$valid_terms = array();
+
+		sort( $available_terms );
+
+		foreach ( $available_terms as $term ) {
+			$monthly_amount = $this->calculate_monthly_payment_amount(
+				$purchase_amount,
+				$term,
+				$this->capabilities()->is_zero_interest() ? 0 : $provider_detail['interest_rate']
```
jonrandy: Wasn't this `is_zero_interest` check already done in `get_available_providers`?

includes/backends/class-omise-backend-installment.php
guzzilar: File: `includes/backends/class-omise-backend-installment.php`
```php
$provider->available_installment_plans = $this-> get_available_instalment_plans(...);

public function get_available_instalment_plans( $allowed_terms, $provider_detail, $purchase_amount ) {
    $plans = array();

    foreach ( $allowed_terms as $term ) {
        $monthly_amount = $this->calculate_monthly_payment_amount( ... );

        $plans[] = array(
            'term'           => $term,
            'monthly_amount' => $monthly_amount
        );
    }

    return $plans;
}
```

`templates/payment/form-installment.php`
```html
<select ...>
    <option>Select term</option>
    <?php foreach ( $backend->available_installment_plans as $installment_plan ) : ?>
        <option value="<?php echo $installment_plan['term']; ?>">
            <?php
            echo sprintf(
                __( '%d months', 'omise', 'omise_installment_term_option' ),
                $installment_plan['term']
            );
            ?>

            <?php
            echo sprintf(
                __( '( %s / months )', 'omise', 'omise_installment_payment_per_month' ),
                wc_price( $installment_plan['monthly_amount'] )
            );
            ?>
        </option>
    <?php endforeach; ?>
</select>
```

@jonrandy LGTY?

includes/backends/class-omise-backend-installment.php
guzzilar: @jonrandy Good point 👍  I think I can clean this one a bit more

includes/backends/class-omise-backend-installment.php
jonrandy: 'term_length' instead of 'term'

guzzilar: @jonrandy 👍 thanks for the input. Checked your rewrite-code. 
Any idea why `interest_rate` is better name than `interest` here?
```php
foreach ( $providers as &$provider ) {
	$provider->provider_code             = str_replace( 'installment_', '', $provider->_id );
	$provider->provider_name             = isset( self::$providers[ $provider->_id ] ) ? self::$providers[ $provider->_id ]['title'] : strtoupper( $provider->code );
	$provider->allowed_installment_plans = $this->get_valid_plans( $provider->allowed_installment_terms, self::$providers[ $provider->_id ], $purchase_amount );
	$provider->interest_rate             = $this->capabilities()->is_zero_interest() ? 0 : ( self::$providers[ $provider->_id ]['interest_rate'] * 100 );
}
```

jonrandy: Because `interest` and `interest_rate` are entirely different things

guzzilar:
@jonrandy Just pushed https://github.com/omise/omise-woocommerce/pull/115/commits/24cb3855ad524b74ab63bfe6563e10ba98ee42ad

Can you help me check one more time?
Thanks

cc @jacstn

guzzilar:
Note, I've run the tests & check at the UI. All works fine to me.

includes/backends/class-omise-backend-installment.php
```diff
@@ -0,0 +1,127 @@
...
+		// Note: as installment payment at the moment only supports for THB currency, so the 
+		//       $purchase_amount is multiplied with 100 to convert the amount into subunit (satang).
+		$providers = $this->capabilities()->getInstallmentBackends( $currency, ( $purchase_amount * 100 ) );
+
+		foreach ( $providers as &$provider ) {
+			$provider_detail = self::$providers[ $provider->_id ];
+
+			$provider->provider_code   = str_replace( 'installment_', '', $provider->_id );
+			$provider->provider_name   = isset( $provider_detail ) ? $provider_detail['title'] : strtoupper( $provider->code );
+			$provider->interest_rate   = $this->capabilities()->is_zero_interest() ? 0 : ( $provider_detail['interest_rate'] * 100 );
```
jonrandy: Why do we have interest rates as a decimal in one place, and suddenly changed to a percentage here?

includes/backends/class-omise-backend-installment.php
```diff
@@ -0,0 +1,127 @@
...
+
+	/**
+	 * @param  float $purchase_amount
+	 * @param  int   $term_length      A length of a given installment term.
+	 * @param  float $interest_rate    its value can be '0' if merchant absorbs an interest.
+	 *
+	 * @return float  of a installment monthly payment (round up to 2 decimals).
+	 */
+	public function calculate_monthly_payment_amount( $purchase_amount, $term_length, $interest_rate ) {
+		$interest = $purchase_amount * $term_length * $interest_rate;
```
jonrandy: This strikes me as wrong. Interest should just be `amount * interest rate` as I understand instalment interest:

If I read that I was going to pay 15% interest on an instalment plan when the cost of the item was say B100, I would expect to pay B115 in total, but it seems here that the interest rate shown is used to calculate the amount of interest paid each and every month. So, if I selected 10 months at 15%, I'd end up paying in total 2.5 times the actual price of the item

jonrandy: I probably used the same formula, but are we sure it is correct?

guzzilar: @jonrandy according to a screenshot that Var have tested with a couple banks (installment provider).
At 3,000 THB, interest rate 0.8%, for 10 months. Net total is 3,240 THB (324 THB per month).
The interest of installment takes the term length into account as well.

So no, in case of 100 THB item with interest rate 15% for 10 months, you are not paying in total 115 THB but 250 THB. The longer you split the bill, the more you pay.

But as your comment:
> Why do we have interest rates as a decimal in one location in the code, and suddenly changed to a percentage here?

I think this can be improved a bit.
I'll update this code to
1. Interest rate as percentage
```php
'installment_bay' => array(
	'bank_code'          => 'bay',
	'title'              => __( 'Krungsri', 'omise' ),
	'interest_rate'      => 0.8, // instead of 0.008
	'min_allowed_amount' => 300.00,
),
```

2. Remove multiply 100 at
```php
$provider->interest_rate      = $this->capabilities()->is_zero_interest() ? 0 : ( $provider_detail['interest_rate'] * 100 );
```

3. Then, here will be 
```php
$interest = $purchase_amount * $interest_rate * $term_length / 100;
```

guzzilar: Says cheese and I'll push the update 🧀

includes/gateway/class-omise-payment-installment.php
```diff
@@ -69,27 +71,15 @@ public function init_form_fields() {
...
-			$currency             = get_woocommerce_currency();
-			$cart_total           = WC()->cart->total;
-			$capabilities         = Omise_Capabilities::retrieve();
-			$installment_backends = $capabilities->getInstallmentBackends( $currency, $this->format_amount_subunit( $cart_total, $currency ) );
-
-			foreach ( $installment_backends as &$backend ) {
-				$backend->provider_code = str_replace( 'installment_', '', $backend->_id );
-				$backend->provider_name = isset( $provider_names[ $backend->_id ] ) ? $provider_names[ $backend->_id ] : strtoupper( $backend->provider_code );
-			}
+			$currency   = get_woocommerce_currency();
```
jonrandy: Curious if this is the correct way? I would expect this to get the general currency setting for the site. In my experience with ecommerce sites, this isn't always correct as sometimes multi-currency plugins are in play. Is the a currency actually stored in the cart - it would probably be a better one to use?

templates/payment/form-installment.php
```diff
@@ -10,15 +10,39 @@
...
+								</span>
+							<?php endif; ?>
 						</div>
 					</label>
 				</li>
 			<?php endforeach; ?>
 		</ul>
+		<div class="omise-buttom-note">
+			<p>
+				<?php echo $viewData['is_zero_interest'] ? 'All installment payments are interest free' : __( 'Monthly payment rates shown may be inaccurate as interest rates are subject to change by its bank issuer.', 'omise' ); ?>
```
jonrandy: Untranslated string

includes/gateway/class-omise-payment-installment.php
guzzilar: I was looking through the whole code of WooCommerce and seems they heavily used `get_woocommerce_currency()` in everywhere. But good point on the multi-currency thing.
I'll do a bit of research and get back to you soon 👍

guzzilar: @jonrandy Checked. I'm convinced that this is the right way (for now), as for 
1. As WooCommerce leaves the hook here, I believe that this is an intention of WC to let those multi-currency feature will hook this filter in order to alternate the currency code.
```php
function get_woocommerce_currency() {
	return apply_filters( 'woocommerce_currency', get_option( 'woocommerce_currency' ) );
}
```

2. I checked how WC formats and displays its price at the checkout page, found that `get_woocommerce_currency()` has been heavily used everywhere.

3. Order has not been created yet at this point. We cannot use a currency from an order object (it will be created at the 'place order' step (aka. payment processing step)).

templates/payment/form-installment.php
```diff
@@ -10,15 +10,39 @@
...
+								</span>
+							<?php endif; ?>
 						</div>
 					</label>
 				</li>
 			<?php endforeach; ?>
 		</ul>
+		<div class="omise-buttom-note">
+			<p>
+				<?php echo $viewData['is_zero_interest'] ? __( 'All installment payments are interest free', 'omise' ) : __( 'Monthly payment rates shown may be inaccurate as interest rates are subject to change by its bank issuer.', 'omise' ); ?>
```
guzzilar: @jonrandy done here 👍

PR #118, opened 2019-07-08
Better way to handle amount subunit

guzzilar:
#### 1. Objective

As increasing of currencies that can be used with Omise Payment, and as a better and a bit more complex of the code that will be used to handle amount subunit converting and to prevent some unexpected situation.

This pull request is aiming to provide a bit of a better place to house those code and logic.

Bonus point is now we can write a unit test to make sure that the converted-amount will be treated correctly before passing to Omise API.

#### 2. Description of change

1. Introducing `Omise_Money` as a dedicated class to handle those converting WC order amount to the subunit format that Omise API requires.

2. Replacing all `Omise_Payment::format_amount_subunit()` with `Omise_Money`.

3. Adding test cases.

#### 3. Quality assurance

**🔧 Environments:**

- **PHP**: v7.3.3
- **WordPress**: v5.2.2
- **WooCommerce**: v3.6.5

**✏️ Details:**

According to the test cases, the unit test script should be green. 
Screen Shot 2562-07-07 at 14 16 15

Screen Shot 2562-07-07 at 14 16 32

Also tested by creating a new charge with Alipay, Credit Card, Installment, Internet Banking to make sure that the order amount is converted properly.

#### 4. Impact of the change

Nothing.

#### 5. Priority of change

Normal.

#### 6. Additional Notes

1. Note, seems now the unit test configuration is broken.
All test cases won't be executed if run the following command:  `vendor/bin/phpunit tests`.
  For now, need to specify the exact test script's path, in order to test each cases.
  ```ssh
  vendor/bin/phpunit tests/unit/includes/class-omise-money-test.php
   
  vendor/bin/phpunit tests/unit/includes/backends/class-omise-backend-installment-test.php
  ```

2. Also note that this pull request doesn't touch on "preventing unsupported currency for each payment method" perspective, but just to improve the way to handle converting WC order amount to Omise subunit amount format.
Some code may still remains and look a bit off (i.e. `is_currency_support` method), but it will be removed / refactored in the next coming PR.

### DIFF
--- a/includes/class-omise-money.php
+++ b/includes/class-omise-money.php
@@ -2,7 +2,7 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
- * @since 3.5
+ * @since 3.6
  */
 class Omise_Money {
 	/**
--- a/includes/class-omise-money.php
+++ b/includes/class-omise-money.php
@@ -48,7 +48,7 @@ class Omise_Money {
 	}
 
 	/**
-	 * @param mixed $amount
+	 * @param  int|float|string $amount
 	 *
 	 * @return int|float|string
 	 */
@@ -72,7 +72,7 @@ class Omise_Money {
 	}
 
 	/**
-	 * @return float
+	 * @return int|float|string  Depending on what type of value that is passed through the construction.
 	 */
 	public function getAmount() {
 		return $this->amount;
--- a/includes/class-omise-money.php
+++ b/includes/class-omise-money.php
@@ -43,16 +43,16 @@ class Omise_Money {
 			throw new Exception( __( 'We do not support the currency you are using.', 'omise' ) );
 		}
 
-		$this->amount   = (float) $this->purifyAmount( $amount );
+		$this->amount   = $this->purify_amount( $amount );
 		$this->currency = strtoupper( $currency );
 	}
 
 	/**
 	 * @param  int|float|string $amount
 	 *
-	 * @return int|float|string
+	 * @return float
 	 */
-	public function purifyAmount( $amount ) {
+	public function purify_amount( $amount ) {
 		if ( ! is_string( $amount ) && ! is_float( $amount ) && ! is_numeric( $amount ) ) {
 			throw new Exception( __( 'An amount has to be integer, float, or string.', 'omise' ) );
 		} 
@@ -61,27 +61,27 @@ class Omise_Money {
 			$amount = preg_replace("/[^0-9.]/", '', $amount);
 		}
 
-		return $amount;
+		return (float) $amount;
 	}
 
 	/**
 	 * @return int
 	 */
-	public function toSubunit() {
+	public function to_subunit() {
 		return (int) ( ( floor( $this->amount * 100 ) / 100 ) * $this->subunit_multiplier[ $this->currency ] );
 	}
 
 	/**
-	 * @return int|float|string  Depending on what type of value that is passed through the construction.
+	 * @return float
 	 */
-	public function getAmount() {
+	public function get_amount() {
 		return $this->amount;
 	}
 
 	/**
 	 * @return string
 	 */
-	public function getCurrency() {
+	public function get_currency() {
 		return $this->currency;
 	}
 }
--- a/includes/class-omise-money.php
+++ b/includes/class-omise-money.php
@@ -28,7 +28,14 @@ class Omise_Money {
 	 * @param  int|float|string $amount
 	 * @param  string           $currency
 	 *
-	 * @return int
+	 * @return int|float  Note that the expected output value's type of this method is to be `int` as Omise Charge API requires.
+	 *                    However, there is a case that this method will return a `float` regarding to
+	 *                    the improper WooCommerce currency setting, which considered as an invalid type of amount.
+	 *
+	 *                    And we would like to let the API raises an error out loud instead of silently remove
+	 *                    or casting a `float` value to `int` subunit.
+	 *                    This is to prevent any miscalculation for those fractional subunits
+	 *                    between the amount that is charged, and the actual amount from the store.
 	 */
 	public function to_subunit( $amount, $currency ) {
 		$amount   = static::purify_amount( $amount );
--- a/includes/class-omise-money.php
+++ b/includes/class-omise-money.php
@@ -25,8 +25,8 @@ class Omise_Money {
 	);
 
 	/**
-	 * @param  int|float $amount
-	 * @param  string    $currency
+	 * @param  int|float|string $amount
+	 * @param  string           $currency
 	 *
 	 * @return int
 	 */
@@ -38,7 +38,7 @@ class Omise_Money {
 			throw new Exception( __( 'We do not support the currency you are using.', 'omise' ) );
 		}
 
-		return (int) ( ( floor( $amount * 100 ) / 100 ) * static::$subunit_multiplier[ $currency ] );
+		return $amount * static::$subunit_multiplier[ $currency ];
 	}
 
 	/**
@@ -48,7 +48,7 @@ class Omise_Money {
 	 */
 	private static function purify_amount( $amount ) {
 		if ( ! is_numeric( $amount ) ) {
-			throw new Exception( __( 'Invalid amount. An amount has to be in a number format.', 'omise' ) );
+			throw new Exception( __( 'Invalid amount type given. Should be int, float, or numeric string.', 'omise' ) );
 		}
 
 		return (float) $amount;
--- a/includes/class-omise-money.php
+++ b/includes/class-omise-money.php
@@ -1,0 +1,87 @@
+<?php
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * @since 3.5
+ */
+class Omise_Money {
+	/**
+	 * @var float
+	 */
+	protected $amount;
+
+	/**
+	 * @var string
+	 */
+	protected $currency;
+
+	/**
+	 * @var array
+	 */
+	private $subunit_multiplier = array(
+		'AUD' => 100,
+		'CAD' => 100,
+		'CHF' => 100,
+		'CNY' => 100,
+		'DKK' => 100,
+		'EUR' => 100,
+		'GBP' => 100,
+		'HKD' => 100,
+		'JPY' => 1,
+		'MYR' => 100,
+		'SGD' => 100,
+		'THB' => 100,
+		'USD' => 100
+	);
+
+	/**
+	 * @param int|float|string $amount
+	 * @param string           $currency
+	 */
+	public function __construct( $amount, $currency ) {
+		if ( ! isset( $this->subunit_multiplier[ strtoupper( $currency ) ] ) ) {
+			throw new Exception( __( 'We do not support the currency you are using.', 'omise' ) );
+		}
+
+		$this->amount   = (float) $this->purifyAmount( $amount );
+		$this->currency = strtoupper( $currency );
+	}
+
+	/**
+	 * @param mixed $amount
+	 *
+	 * @return int|float|string
+	 */
+	public function purifyAmount( $amount ) {
+		if ( ! is_string( $amount ) && ! is_float( $amount ) && ! is_numeric( $amount ) ) {
+			throw new Exception( __( 'An amount has to be integer, float, or string.', 'omise' ) );
+		} 
+
+		if ( is_string( $amount ) ) {
+			$amount = preg_replace("/[^0-9.]/", '', $amount);
+		}
+
+		return $amount;
+	}
+
+	/**
+	 * @return int
+	 */
+	public function toSubunit() {
+		return (int) ( ( floor( $this->amount * 100 ) / 100 ) * $this->subunit_multiplier[ $this->currency ] );
+	}
+
+	/**
+	 * @return float
+	 */
+	public function getAmount() {
+		return $this->amount;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getCurrency() {
+		return $this->currency;
+	}
+}
--- a/includes/class-omise-money.php
+++ b/includes/class-omise-money.php
@@ -26,19 +26,6 @@ class Omise_Money {
 
 	/**
 	 * @param  int|float $amount
-	 *
-	 * @return float
-	 */
-	public static function purify_amount( $amount ) {
-		if ( ! is_numeric( $amount ) ) {
-			throw new Exception( __( 'An amount has to be integer, or float.', 'omise' ) );
-		}
-
-		return (float) $amount;
-	}
-
-	/**
-	 * @param  int|float $amount
 	 * @param  string    $currency
 	 *
 	 * @return int
@@ -52,5 +39,18 @@ class Omise_Money {
 		}
 
 		return (int) ( ( floor( $amount * 100 ) / 100 ) * static::$subunit_multiplier[ $currency ] );
+	}
+
+	/**
+	 * @param  int|float $amount
+	 *
+	 * @return float
+	 */
+	private static function purify_amount( $amount ) {
+		if ( ! is_numeric( $amount ) ) {
+			throw new Exception( __( 'Invalid amount. An amount has to be in a number format.', 'omise' ) );
+		}
+
+		return (float) $amount;
 	}
 }
--- a/includes/class-omise-money.php
+++ b/includes/class-omise-money.php
@@ -6,19 +6,9 @@ defined( 'ABSPATH' ) || exit;
  */
 class Omise_Money {
 	/**
-	 * @var float
-	 */
-	protected $amount;
-
-	/**
-	 * @var string
-	 */
-	protected $currency;
-
-	/**
 	 * @var array
 	 */
-	private $subunit_multiplier = array(
+	private static $subunit_multiplier = array(
 		'AUD' => 100,
 		'CAD' => 100,
 		'CHF' => 100,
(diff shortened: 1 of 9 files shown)


includes/class-omise-money.php
```diff
@@ -0,0 +1,87 @@
...
+	 * @param int|float|string $amount
+	 * @param string           $currency
+	 */
+	public function __construct( $amount, $currency ) {
+		if ( ! isset( $this->subunit_multiplier[ strtoupper( $currency ) ] ) ) {
+			throw new Exception( __( 'We do not support the currency you are using.', 'omise' ) );
+		}
+
+		$this->amount   = (float) $this->purifyAmount( $amount );
+		$this->currency = strtoupper( $currency );
```
guzzilar: Note, the part that I am still 50/50 is that, should we transform a currency string that is passed at the constructor to an upper case or better just leave it as it is (whatever, lower-case or upper-case).

All opinions are welcomed.

jonrandy:
Will review properly later, but first thought is that this is something that really belongs in `OmisePHP` since otherwise it will effectively be duplicated in every plugin

guzzilar:
> Will review properly later, but first thought is that this is something that really belongs in `OmisePHP` since otherwise it will effectively be duplicated in every plugin

@jonrandy maybe, it's kinda always 50/50 to me all this time either should we keep the lib lean and do only simple job, as to connect with Omise API. Or adding this sort of helper class and method.

However, I kinda want to freeze OmisePHP at the moment til we finish the design of OmisePHP v3.0. Keep adding stuff and it will be forever hard to complete the new version. And probably we can start adding this afterward and it wouldn't give an odd feeling. 

Thanks for the input btw. 👍

guzzilar:
Ah, I just found a test where the test is right but the feeling is off, @jacstn @jonrandy.
Let me update the code a bit and will ping you again. 🙏

jonrandy:
I think when working with OmisePHP to make charges, converting currency amounts to subunit amounts is something that is always going to be necessary to do - both in our plugins, and in 3rd party code using it. I think this is a definite candidate for OmisePHP functionality. It will allow us to have a single, centralised, standard way to do this conversion - removing the possibility of slip-ups or flawed implementations done outside

guzzilar:
Done @jacstn @jonrandy

includes/class-omise-money.php
```diff
@@ -0,0 +1,87 @@
...
+
+	/**
+	 * @var string
+	 */
+	protected $currency;
+
+	/**
+	 * @var array
+	 */
+	private $subunit_multiplier = array(
```
jonrandy: Would it not be easier just to store ones that aren't 100? and if not in the list, just default to 100?

jonrandy: Currency support is handled elsewhere. I don't think this code should be dealing with that

includes/class-omise-money.php
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	/**
+	 * @param  int|float|string $amount
+	 *
+	 * @return float
+	 */
+	public function purify_amount( $amount ) {
+		if ( ! is_string( $amount ) && ! is_float( $amount ) && ! is_numeric( $amount ) ) {
+			throw new Exception( __( 'An amount has to be integer, float, or string.', 'omise' ) );
```
jonrandy: Seems a little OTT supporting strings and formatted strings

includes/class-omise-money.php
```diff
@@ -0,0 +1,87 @@
...
+			$amount = preg_replace("/[^0-9.]/", '', $amount);
+		}
+
+		return (float) $amount;
+	}
+
+	/**
+	 * @return int
+	 */
+	public function to_subunit() {
```
jonrandy: No subunit to amount function?

jonrandy: One example of where this is needed is showing the user limitations of instalment payments (from capabilities) - which are expressed in subunits

includes/class-omise-money.php
guzzilar: > Would it not be easier just to store ones that aren't 100? and if not in the list, just default to 100?
This method could be simplify to 
```php
$subunit_multiplier = [ 'jpy' => 1 ];
$multiplier         = isset( $subunit_multiplier[ $order->currency ] ) ? $subunit_multiplier[ $order->currency ] : 100;
```

I chose another way around as showing all currencies is just more clearer to read (might not the shortest and smartest code, but I account for the readability on this one).

However, there might be another way I didn't think of. Wanna share some snippet of code of your thought?

includes/class-omise-money.php
guzzilar: I was thinking about a situation where we are not sure if the platform we are integrating with is returning an amount in a string, float, or int type – which is pretty common in PHP...

What do you say? Better to remove the support of string? or you just wanna leave a note only?
(to me, I prefer to keep it, but you may wanna explain more the downside of this that I didn't see)

includes/class-omise-money.php
jonrandy: Pretty much what you showed above. It is much simpler, and reduces future maintenance burden (it will rarely need to change when new currencies are added)

includes/class-omise-money.php
guzzilar: > No subunit to amount function? 
there is `getAmount()` which returns a original amount that is passed through this class.

> One example of where this is needed is showing the user limitations of instalment payments (from capabilities) - which are expressed in subunits
Sorry, can you explain a bit more? Not sure if I get it.

guzzilar:
> Overall, I feel this is overly complex. It also has some missing functionality. I think it would be better implemented in a simple utility class with static functions. Creating an object every time this functionality is needed seems wasteful and unnecessary

@jonrandy  Good point thou, I might have a bit different opinion on this but so far, there is still no concrete case to use `Omise_Money` object, so probably I'll try to turn `to_subunit()` method into a static one, and just leave the rest as it is.

Btw, can you point me out a bit on `It also has some missing functionality`. What is missing?
For _"No subunit to amount function?"_, yes, there is `getAmount()` method.

includes/class-omise-money.php
jonrandy: I just seems like overkill. Especially when this code is being added to a known environment (our woo_commerce plugin) - where we know what to expect. Unless you are advocating for this to be a separate class that we also add in to other projects - in which case we would be better off adding it to `omise-php` as discussed. I really think though, that in the case of units/subunits, all we need is 2 simple static methods on some kind of Omise utility class

includes/class-omise-money.php
jonrandy: If a limit for a certain payment type is given to us in subunits, we may obviously need to present this to the user as a readable amount. For this, we obviously need to be able to convert from subunits to amount

includes/class-omise-money.php
guzzilar: I agreed on the simplicity. But still, wanna stand for the readability in this case.

The above one could be easy to maintain, but raising more questions at the same time.
It is not just benefit us but for anyone that read the code without context of what currency Omise supported.

How about just leave it as it is for now, and I will change to the short one later if it's really a burden for maintenance?

includes/class-omise-money.php
guzzilar: Right, I was designing this method with a thought of "this will be used all over projects".
Well, agreed, `An amount has to be string` seems weird. Removing

includes/class-omise-money.php
jonrandy: The currencies that Omise support aren't really relevant to what should be very simple conversion functions - which is all that we need here

includes/class-omise-money.php
guzzilar: ic. Didn't think of the case converting another way around (Omise API to frontend).
I feel a bit off if "converting Omise API to readable format" method have to be implemented here (as it wasn't designed for).

Let me think a bit. 
In the meantime, any suggestion? (I'm trying to avoid such as `Omise_Utility`, `Omise_Helper` classes)

includes/class-omise-money.php
guzzilar: I understand your point of view and idea for the simplicity.
To me, it's just choosing between explicit way or smart way and I normally tend to go with an explicit way if possible.

Is there any concrete benefit to switch to a smarter way? (besides maintenance – automatically support for a new currency that multiply 100 for its subunit by default – which I will be willingly to change if it comes to really a burden for a maintenance later on),

jonrandy:
This would be my first crack at a solution

```php
class OmiseUtils {
	
	
	private static
		$currencySubunits = [
			"JPY" => 1
		],
		$defaultSubunits = 100
	;
	
	public static function amountToSubunits($amount, $curr) {
		return $amount * self::subUnitsFor($curr);
	}
	
	public static function subunitsToAmount($subUnits, $curr) {
		return $subUnits / self::subUnitsFor($curr);
	}
	
	public static function subUnitsFor($curr) {
		$c = strtoupper($curr);
		return isset(self::$currencySubunits[$c]) ? self::$currencySubunits[$c] : self::$defaultSubunits;
	}
	
}
```

includes/class-omise-money.php
guzzilar: (many parts of the project could be re-written into just 1 or 2 lines and it would function just the same, yet, we break the line, we add a meaningless variable and assign a value to it just we can pass it to another function at the very next line, etc. I account that for readability sake – same as this code)

guzzilar:
@jonrandy I buy for 
```php
public static function amountToSubunits($amount, $curr) {
    return $amount * self::subUnitsFor($curr);
}

public static function subunitsToAmount($subUnits, $curr) {
    return $subUnits / self::subUnitsFor($curr);
}

public static function subUnitsFor($curr) {
    $c = strtoupper($curr);
    return isset(self::$currencySubunits[$c]) ? self::$currencySubunits[$c] : self::$defaultSubunits;
}
```

Any better name than `Utils`? I'm trying to avoid those `Utils` and `Helpers` name as possible as it will turn into a place where we dump whatever we don't know where it should belong kind of habit...

jonrandy:
Cache invalidation and naming things...

:)

includes/class-omise-money.php
```diff
@@ -0,0 +1,56 @@
...
+	 * @param  int|float $amount
+	 * @param  string    $currency
+	 *
+	 * @return int
+	 */
+	public function to_subunit( $amount, $currency ) {
+		$amount   = static::purify_amount( $amount );
+		$currency = strtoupper( $currency );
+
+		if ( ! isset( static::$subunit_multiplier[ $currency ] ) ) {
```
jonrandy: I'm against having any currency checks here. This is handled elsewhere and is unnecessary here. Adding it here reduces the function's utility

includes/class-omise-money.php
```diff
@@ -0,0 +1,56 @@
...
+		'THB' => 100,
+		'USD' => 100
+	);
+
+	/**
+	 * @param  int|float $amount
+	 *
+	 * @return float
+	 */
+	public static function purify_amount( $amount ) {
```
jonrandy: Does this need to be public?

includes/class-omise-money.php
```diff
@@ -0,0 +1,56 @@
...
+	public function to_subunit( $amount, $currency ) {
+		$amount   = static::purify_amount( $amount );
+		$currency = strtoupper( $currency );
+
+		if ( ! isset( static::$subunit_multiplier[ $currency ] ) ) {
+			throw new Exception( __( 'We do not support the currency you are using.', 'omise' ) );
+		}
+
+		return (int) ( ( floor( $amount * 100 ) / 100 ) * static::$subunit_multiplier[ $currency ] );
+	}
```
jonrandy: Still no `subunit_to_amount`? This is necessary in places and this class would seem to be the correct home

includes/class-omise-money.php
jonrandy: I'm not sure this function is even necessary, but if we really must have it, we need to correct the exception message as this function is perfectly happy accepting numeric strings. i.e. `purify_amount("123")` works just fine

includes/class-omise-money.php
guzzilar: Unnecessary is understandable, but why do you think it will reduce the function's utility?
Isn't that utility should not rely on any outside its context (aka. class) to handle something for them? Or else the code will be dependent with that particular code structure?

Also, I think the currency check purpose here is different from ones in each payment method classes.
Those in each payment method classes are to check if the currencies are applicable for a specific method. While here is to check if a given currency can be converted to subunit.

includes/class-omise-money.php
guzzilar: At the moment, there is no place that needs `subunit_to_amount`. So I didn't have it here, for now.

includes/class-omise-money.php
guzzilar: @jonrandy any idea for the exception message? `An amount has to be number`?

guzzilar:
@jonrandy Changed `purify_amount` visibility to `private` & updated the exception message.
For the rest of your feedbacks, I've commented back already. Let me know if you have any more concern left, thanks 👍

includes/class-omise-money.php
jonrandy: It will reduce the function's utility as without the restriction, the function can handle many many more cuurencies. However, a handful of currencies would be wrong. If we switch to this:

```php
class OmiseUtils {
	
	private static
		$currencyDecimals = [
			"BIF" => 0,
			"CLP" => 0,
			"DJF" => 0,
			"GNF" => 0,
			"ISK" => 0,
			"JPY" => 0,
			"KMF" => 0,
			"KRW" => 0,
			"PYG" => 0,
			"RWF" => 0,
			"UGX" => 0,
			"UYI" => 0,
			"VND" => 0,
			"VUV" => 0,
			"XAF" => 0,
			"XOF" => 0,
			"XPF" => 0,
			"BHD" => 3,
			"IQD" => 3,
			"JOD" => 3,
			"KWD" => 3,
			"LYD" => 3,
			"OMR" => 3,
			"TND" => 3,
		],
		$defaultCurrencyDecimals = 2
	;
	
	public static function amountToSubunits($amount, $curr) {
		return $amount * self::subUnitsFor($curr);
	}
	
	public static function subunitsToAmount($subUnits, $curr) {
		return $subUnits / self::subUnitsFor($curr);
	}
	
	public static function subUnitsFor($curr) {
		$c = strtoupper($curr);
		$decimals = isset(self::$currencyDecimals[$c]) ? self::$currencyDecimals[$c] : self::$defaultCurrencyDecimals;
		return pow(10, $decimals);
	}
	
}
```

We cover all currently used currencies worldwide. I think this class should be purely about money/currency and not be bound in any way by any restrictions Omise has. They really don't belong here

includes/class-omise-money.php
```diff
@@ -0,0 +1,56 @@
...
+
+	/**
+	 * @param  int|float $amount
+	 *
+	 * @return float
+	 */
+	private static function purify_amount( $amount ) {
+		if ( ! is_numeric( $amount ) ) {
+			throw new Exception( __( 'Invalid amount. An amount has to be in a number format.', 'omise' ) );
```
jonrandy: Maybe "Invalid amount type given. Should be int, float, or numeric string"

includes/class-omise-money.php
```diff
@@ -0,0 +1,56 @@
...
+		'HKD' => 100,
+		'JPY' => 1,
+		'MYR' => 100,
+		'SGD' => 100,
+		'THB' => 100,
+		'USD' => 100
+	);
+
+	/**
+	 * @param  int|float $amount
```
jonrandy: string?

includes/class-omise-money.php
```diff
@@ -0,0 +1,56 @@
...
+	 */
+	public function to_subunit( $amount, $currency ) {
+		$amount   = static::purify_amount( $amount );
+		$currency = strtoupper( $currency );
+
+		if ( ! isset( static::$subunit_multiplier[ $currency ] ) ) {
+			throw new Exception( __( 'We do not support the currency you are using.', 'omise' ) );
+		}
+
+		return (int) ( ( floor( $amount * 100 ) / 100 ) * static::$subunit_multiplier[ $currency ] );
```
jonrandy: I don't think we should do any rounding/removing of additional decimal places. This could result in amounts that have rounding errors being charged with no warning this was happening

includes/class-omise-money.php
guzzilar: Isn't it over-design?

includes/class-omise-money.php
guzzilar: I was trying to cover as much as weird case as possible. 
Can you check my test cases and let me know if there is any potential edge cases that I may missed, I will add more test case afterward?

includes/class-omise-money.php
jonrandy: I think it's actually a simpler design than you propose (i.e. not listing detail for every currency) , and can do more

jonrandy: The decimals thing was just to reduce having to type loads of zeros repeatedly

includes/class-omise-money.php
guzzilar: 👍  adding

includes/class-omise-money.php
```diff
@@ -0,0 +1,56 @@
...
+		'SGD' => 100,
+		'THB' => 100,
+		'USD' => 100
+	);
+
+	/**
+	 * @param  int|float $amount
+	 * @param  string    $currency
+	 *
+	 * @return int
```
jonrandy: I think we should return fractional subunits if their input results in this. We shouldn't quietly be removing money. Whilst this is almost insignificant on a single conversion, repeated conversions (I can't think of a good example scenario - maybe micropayments?) over time could result in considerable discrepancies

jonrandy: If we remove stuff quietly like this, the developer may not realise their initial mistake as it is being swept under the carpet. Better to let them get a nasty error when they try to charge with the fractional subunits

guzzilar:
@jonrandy agreed with "should not remove those fractional subunits secretly".
I updated the code. Instead, just let OmiseCharge raises an error when those fractional subunits are included in the amount parameter that use to create a charge.

includes/class-omise-money.php
```diff
@@ -0,0 +1,56 @@
...
+		'SGD' => 100,
+		'THB' => 100,
+		'USD' => 100
+	);
+
+	/**
+	 * @param  int|float|string $amount
+	 * @param  string           $currency
+	 *
+	 * @return int
```
jonrandy: int|float ?

includes/class-omise-money.php
```diff
@@ -0,0 +1,56 @@
...
+	 */
+	public function to_subunit( $amount, $currency ) {
+		$amount   = static::purify_amount( $amount );
+		$currency = strtoupper( $currency );
+
+		if ( ! isset( static::$subunit_multiplier[ $currency ] ) ) {
```
jonrandy: If we're insisting on keeping the currency check (which I still think really shouldn't be here) then we should really split it out into a `currency_supported` function? As it will also be used in `subunit_to_amount`

guzzilar: I would like to consider having that after we really implement that `subunit_to_amount` method (It's a good practice to reduce the duplicated code, but not yet to happened at the moment).

includes/class-omise-money.php
guzzilar: geez, miss this one. Thanks 👍

PR #120, opened 2019-07-22
Properly assigning querystrings for the payment callback endpoints

guzzilar:
#### 1. Objective

The current code is using a redundant approach to assign a querystring to the payment callback endpoints.

```php
add_query_arg( 'order_id', $order_id, site_url() . "?wc-api=omise_alipay_callback" );
```

See that we use a function provided by WordPress, `add_query_arg()` to create a querystring for `order_id` but manually set a querystring for `wc-api` argument: `?wc-api=omise_alipay_callback`.

#### 2. Description of change

Assigning querystrings with a proper way that WordPress has provided.

```php
$return_uri = add_query_arg(
	array(
		'wc-api'   => 'omise_alipay_callback',
		'order_id' => $order_id
	),
	home_url()
);
```

This is to avoid any conflict in case if `home_url` provides any of itself a querystring.

Also, `site_url()` has been replaced with `home_url()` as `site_url` just purely returns a `base` url of an accessible domain name, while `home_url` will be returning whatever url that is set as a `homepage` for WordPress setting.

This likely is to eliminating a potential incompatible URL that may have been hooked by other 3rd-party plugins. 

#### 3. Quality assurance

**🔧 Environments:**

- **WordPress**: v5.2.2.
- **WooCommerce**: v3.6.5.
- **PHP**: v7.3.3.

**✏️ Details:**

Just test create a charge with Alipay, Credit Card, Installment, Internet Banking payments to see if the `return_uri` is properly set the same as before.

Alipay
Screen Shot 2562-07-22 at 21 39 19

Credit Card
Screen Shot 2562-07-22 at 21 36 11

Installment
Screen Shot 2562-07-22 at 21 37 40

Internet Banking
Screen Shot 2562-07-22 at 21 34 53

#### 4. Impact of the change

Nothing

#### 5. Priority of change

Normal

#### 6. Additional Notes

Nothing

### DIFF
--- a/includes/gateway/class-omise-payment-alipay.php
+++ b/includes/gateway/class-omise-payment-alipay.php
@@ -68,13 +68,20 @@ function register_omise_alipay() {
 				apply_filters( 'omise_charge_params_metadata', array(), $order ),
 				array( 'order_id' => $order_id ) // override order_id as a reference for webhook handlers.
 			);
+			$return_uri = add_query_arg(
+				array(
+					'wc-api'   => 'omise_alipay_callback',
+					'order_id' => $order_id
+				),
+				home_url()
+			);
 
 			return OmiseCharge::create( array(
 				'amount'      => $this->format_amount_subunit( $order->get_total(), $order->get_order_currency() ),
 				'currency'    => $order->get_order_currency(),
 				'description' => apply_filters( 'omise_charge_params_description', 'WooCommerce Order id ' . $order_id, $order ),
 				'source'      => array( 'type' => 'alipay' ),
-				'return_uri'  => add_query_arg( 'order_id', $order_id, site_url() . "?wc-api=omise_alipay_callback" ),
+				'return_uri'  => $return_uri,
 				'metadata'    => $metadata
 			) );
 		}
--- a/includes/gateway/class-omise-payment-creditcard.php
+++ b/includes/gateway/class-omise-payment-creditcard.php
@@ -244,12 +244,19 @@ function register_omise_creditcard() {
 				}
 			}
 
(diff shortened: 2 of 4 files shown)


includes/gateway/class-omise-payment-alipay.php
```diff
@@ -68,13 +68,20 @@ public function charge( $order_id, $order ) {
 				apply_filters( 'omise_charge_params_metadata', array(), $order ),
 				array( 'order_id' => $order_id ) // override order_id as a reference for webhook handlers.
 			);
+			$return_uri = add_query_arg(
```
jonrandy: Feels like maybe we should have a method for building the `return_uri` in `Omise_Payment` class

jacstn: @guzzilar consider comment above.. I am approving, not a big deal and it is urgent.
tested manually according to your testing description, works fine.

guzzilar: @jonrandy not sure when but I may reinvent the way to handle these callbacks later on after Bill Payment (or after TrueMoney Walltet).
To me, there are some parts that the code is redundant, moving out only `return_uri` without designing the whole rest won't be that so effective.

jonrandy: Yep, I'd noticed other functionality that could move to the parent class too, but was trying to limit the scope of my comments to the changes in this PR

guzzilar: ic. noted and thanks 👍

guzzilar:
Thanks guys, that's fast review :D

PR #122, opened 2019-07-27
Introducing Bill Payment, Tesco Lotus (only available in Thailand)

guzzilar:
#### 1. Objective

As the statement Omise put at the releasing article: 

Our focus is to help you boost sales by offering payment solutions that put customers choice and experience ahead. We’re super excited to let you know that starting today, Omise merchants in Thailand will be able to offer cash payments at Tesco Lotus as an option for customers to pay.

Increasingly we see consumers shifting over to online payments, however a large percentage still prefer to pay by cash over-the-counter. Tesco Lotus is currently the most preferred outlet by consumers, with more than 1800 locations across the country including 24-hour Lotus Express stores.

The objective of this pull request is to bring this feature to our lovely WooCommerce.

https://www.omise.co/now-supporting-over-the-counter-payments-at-tesco-lotus

#### 2. Description of change

Adding all the needs in order to integrate Omise Bill Payment to WooCommerce.
- Implementing all the foundation code to list the payment at WooCommerce payment setting page.
- Displaying Bill Payment payment method at Omise setting page.
- Integrating Bill Payment to WooCommerce checkout page
- Display Bill Payment's barcode at the "order-received" page.

#### 3. Quality assurance

**🔧 Environments:**

- **WordPress**: v5.2.2 (latest)
- **WooCommerce**: v3.6.5 (latest)
- **PHP version**: v7.3.3

**✏️ Details:**

After installed this patch, you will be able to use Omise Bill Payment Tesco.
The below quality assurance steps is to make sure that all functions are properly work as expected.

1. At Omise Payment Setting page (logged in to WordPress admin page first), there will be a new payment method, **"Bill Payment: Tesco"**.
billpayment-setting

2. Once the payment is enabled, new payment method will be displayed at the checkout page.
billpayment-checkout

3. Bill Payment uses the "offline" flow, meaning that there is no redirect to any 3rd party page but create a new charge with `authorize: pending`, and `paid: pending` status then redirect buyer to the order-received page.
billpayment-completed

billpayment-charge

billpayment-log

4. At WooCommerce Order Admin page, a new order will be created with `pending payment` status.
Screen Shot 2562-07-31 at 13 56 40

**IMPORTANT**

5. If Webhook is set, Omise-WooCommerce will catch a `charge.complete` event and update a pending-payment order accordingly. If no, then merchants have to update an order status manually.

Successful case, order status get updated to `processing`.
billpayment-order-paid

Failure case, order status get updated to `failed`.
billpayment-order-failed

[...]

PR #125, opened 2019-08-05
Bill Payment: Displaying Tesco barcode at WC Order Confirmation email.

guzzilar:
> ⚠️ THIS PULL REQUEST REQUIRES #122 TO BE MERGED FIRST.

#### 1. Objective

As for the actual use case, users need to carry this barcode to pay at Tesco counter service.
Regarding to the use case above, this pull request is providing and displaying barcode to WooCommerce Order Confirmation email.

#### 2. Description of change

Adding Tesco barcode to WooCommerce Order Confirmation email by hook WooCommerce's `woocommerce_email_after_order_table` action.

#### 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: v3.6.5
- **WordPress**: v5.2.2
- **PHP**: v7.2.20

**✏️ Details:**

Make a purchase using `Bill Payment: Tesco` payment method as normal.
Screen Shot 2562-08-05 at 10 41 53

After completed the purchase, a Tesco's barcode should be shown at the "Thank-You" page.
Screen Shot 2562-08-05 at 10 42 16

A Tesco's barcode should be shown at the "Order Confirmation" email as well.
Screen Shot 2562-08-05 at 10 44 19

Use a mobile application to scan those 2 barcodes that are shown at the "Thank-You" page and at the "Order Confirmation" email to make sure that the barcode is showing the same information as the below referencing number.
IMG_1834

_note, the application above is showing just only the first part of the barcode but I checked already, all the numbers between the barcode and the reference number are matched._

#### 4. Impact of the change

Nothing

#### 5. Priority of change

Normal

#### 6. Additional Notes

Note that the barcode is worked perfectly on a large screen but we better do an actual test by physically pay the bill at Tesco counter service.

– Thanks @jacstn for the SVG > HTML converter code :D

### DIFF
--- a/includes/gateway/class-omise-payment-billpayment-tesco.php
+++ b/includes/gateway/class-omise-payment-billpayment-tesco.php
@@ -199,7 +199,7 @@ function register_omise_billpayment_tesco() {
 
 				// Set HTML attributes based on <rect> node's attributes.
 				$divRect = $xhtml->createElement( 'div' );
-				$divRect->setAttribute( 'style', "float:left; position:relative; height:50px; width:$width; background-color:#000; margin-left:$margin" );
+				$divRect->setAttribute( 'style', "float: left; position: relative; height: 50px; border-left: $width solid #000000; width: 0; margin-left: $margin" );
 				$xhtml->appendChild( $divRect );
 
 				$prevX     = $attributes['x'];
--- a/includes/gateway/class-omise-payment-billpayment-tesco.php
+++ b/includes/gateway/class-omise-payment-billpayment-tesco.php
@@ -114,24 +114,26 @@ function register_omise_billpayment_tesco() {
 		 * @param string       $context  pass 'email' value through this argument only for 'sending out an email' case.
 		 */
 		public function display_barcode( $order, $context = 'view' ) {
-			if ( ! $order = $this->load_order( $order ) ) {
+			if ( ! $this->load_order( $order ) ) {
 				return;
 			}
 
 			$charge_id    = $this->get_charge_id_from_order();
 			$charge       = OmiseCharge::retrieve( $charge_id );
 			$barcode_svg  = file_get_contents( $charge['source']['references']['barcode'] );
-			$barcode_html = $this->barcode_svg_to_html( $barcode_svg );
 			?>
 
 			<div class="omise omise-billpayment-tesco-details" <?php echo 'email' === $context ? 'style="margin-bottom: 4em; text-align:center;"' : ''; ?>>
 				<p><?php echo __( 'Use this barcode to pay at Tesco Lotus.', 'omise' ); ?></p>
 				<div class="omise-billpayment-tesco-barcode">
-					<?php if ( 'email' === $context ) : ?>
-						<?php echo $barcode_html; ?>
-					<?php else : ?>
-						<?php echo $barcode_svg; ?>
-					<?php endif; ?>
+					<?php
+					if ( 'email' === $context ) {
+						$barcode_html = $this->barcode_svg_to_html( $barcode_svg );
+						echo $barcode_html;
+					} else {
+						echo $barcode_svg;
+					}
+					?>
 				</div>
 				<small class="omise-billpayment-tesco-reference-number">
 					<?php
@@ -148,12 +150,36 @@ function register_omise_billpayment_tesco() {
 			<?php
 		}
 
+		/**
+		 * @param WC_Order $order
+		 *
+		 * @see   woocommerce/templates/emails/email-order-details.php
+		 * @see   woocommerce/templates/emails/plain/email-order-details.php
+		 */
 		public function email_barcode( $order ) {
 			$this->display_barcode( $order, 'email' );
 		}
 
 		/**
 		 * Convert a given SVG Bill Payment Tesco's barcode to HTML format.
+		 *
+		 * Note that the SVG barcode contains with the following structure:
+		 *
+		 * <?xml version="1.0" encoding="UTF-8"?>
+		 * <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="515px" height="90px" viewBox="0 0 515 90" version="1.1" preserveAspectRatio="none">
+		 *   <title>** reference number **</title>
+		 *   <g id="canvas">
+		 *     <rect x="0" y="0" width="515px" height="90px" fill="#fff" />
+		 *     <g id="barcode" fill="#000">
+		 *       <rect x="20" y="20" width="2px" height="50px" />
+		 *       ... (repeat <rect> node for displaying barcode) ...
+		 *       <rect x="493" y="20" width="2px" height="50px" />
+		 *     </g>
+		 *   </g>
+		 * </svg>
+		 *
+		 * The following code in this method is to read all <rect> nodes' attributes under the <g id="barcode"></g>
+		 * in order to replicate the barcode in HTML <div></div> element.
 		 *
 		 * @param  string $barcode_svg
 		 *
@@ -165,13 +191,13 @@ function register_omise_billpayment_tesco() {
 			$prevX     = 0;
 			$prevWidth = 0;
 
-			// Get data from all <rect> nodes.
+			// Read data from all <rect> nodes.
 			foreach ( $xml->g->g->children() as $rect ) {
 				$attributes = $rect->attributes();
 				$width      = $attributes['width'];
 				$margin     = ( $attributes['x'] - $prevX - $prevWidth ) . 'px';
 
-				//set html attributes based on SVG attributes
+				// Set HTML attributes based on <rect> node's attributes.
 				$divRect = $xhtml->createElement( 'div' );
 				$divRect->setAttribute( 'style', "float:left; position:relative; height:50px; width:$width; background-color:#000; margin-left:$margin" );
 				$xhtml->appendChild( $divRect );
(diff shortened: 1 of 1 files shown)


jacstn:
@guzzilar what is the best way to send email from your local machine in WordPress?
does it send it without any additional components?
on Magento 2 you need to install an extra plugin for that.

guzzilar:
@jacstn need an extra plugin for WordPress too. Some SMTP plugin would work.
i.e. https://wordpress.org/plugins/wp-smtp (note, only use this one for test, seems there are some reviews saying this plugin saving data in plain text).

For me, I was deploying this code to my actual server to test.

guzzilar:
Note, just pushed one more commit to update DocBlock and clean up code.

includes/gateway/class-omise-payment-billpayment-tesco.php
```diff
@@ -0,0 +1,226 @@
...
+					'title'   => __( 'Enable/Disable', 'omise' ),
+					'type'    => 'checkbox',
+					'label'   => __( 'Enable Omise Tesco Bill Payment', 'omise' ),
+					'default' => 'no'
+				),
+
+				'title' => array(
+					'title'       => __( 'Title', 'omise' ),
+					'type'        => 'text',
+					'description' => __( 'This controls the title which the user sees during checkout.', 'omise' ),
```
jonrandy: No 'which'

includes/gateway/class-omise-payment-billpayment-tesco.php
```diff
@@ -0,0 +1,226 @@
...
+					'default'     => __( 'Bill Payment: Tesco', 'omise' ),
+				),
+
+				'description' => array(
+					'title'       => __( 'Description', 'omise' ),
+					'type'        => 'textarea',
+					'description' => __( 'This controls the description which the user sees during checkout.', 'omise' )
```
jonrandy: No 'which'

includes/gateway/class-omise-payment-billpayment-tesco.php
```diff
@@ -0,0 +1,226 @@
...
+		/**
+		 * @inheritdoc
+		 */
+		public function result( $order_id, $order, $charge ) {
+			if ( 'failed' == $charge['status'] ) {
+				return $this->payment_failed( $charge['failure_message'] . ' (code: ' . $charge['failure_code'] . ')' );
+			}
+
+			if ( 'pending' == $charge['status'] ) {
+				return [
```
jonrandy: Mixed array styles?

includes/gateway/class-omise-payment-billpayment-tesco.php
```diff
@@ -0,0 +1,226 @@
...
+			if ( ! $this->load_order( $order ) ) {
+				return;
+			}
+
+			$charge_id    = $this->get_charge_id_from_order();
+			$charge       = OmiseCharge::retrieve( $charge_id );
+			$barcode_svg  = file_get_contents( $charge['source']['references']['barcode'] );
+			?>
+
+			<div class="omise omise-billpayment-tesco-details" <?php echo 'email' === $context ? 'style="margin-bottom: 4em; text-align:center;"' : ''; ?>>
```
jonrandy: Is this the eecommended way to add HTML into a page in WC??

includes/gateway/class-omise-payment-billpayment-tesco.php
```diff
@@ -0,0 +1,226 @@
...
+				<p><?php echo __( 'Use this barcode to pay at Tesco Lotus.', 'omise' ); ?></p>
+				<div class="omise-billpayment-tesco-barcode">
+					<?php
+					if ( 'email' === $context ) {
```
jonrandy: Do the emails have plaintext version too? Just wondering what happens if the email was opened on a non-html client (however rare that might be). Do we include a link to the barcode?

includes/gateway/class-omise-payment-billpayment-tesco.php
```diff
@@ -0,0 +1,226 @@
...
+
+			// Read data from all <rect> nodes.
+			foreach ( $xml->g->g->children() as $rect ) {
+				$attributes = $rect->attributes();
+				$width      = $attributes['width'];
+				$margin     = ( $attributes['x'] - $prevX - $prevWidth ) . 'px';
+
+				// Set HTML attributes based on <rect> node's attributes.
+				$divRect = $xhtml->createElement( 'div' );
+				$divRect->setAttribute( 'style', "float:left; position:relative; height:50px; width:$width; background-color:#000; margin-left:$margin" );
```
jonrandy: If this is printed with 'backgrounds' switched off (as is sometimes a common option on printer settings) - will the barcode still show?

includes/gateway/class-omise-payment-billpayment-tesco.php
guzzilar: @jonrandy yup, there is a plaintext email version too in WC email setting.
Can you name me some non-html client that we should test? no idea on that.

includes/gateway/class-omise-payment-billpayment-tesco.php
guzzilar: The "background switched off" printing mode will cancel the css `background`?
If so then no, the barcode will just show empty white space.

Any suggestion?

includes/gateway/class-omise-payment-billpayment-tesco.php
jonrandy: I know Mozilla Thunderbird can be made to open emails in plain text by default

jonrandy: Small article on plain text emails - https://litmus.com/blog/best-practices-for-plain-text-emails-a-look-at-why-theyre-important

includes/gateway/class-omise-payment-billpayment-tesco.php
jonrandy: Maybe make the 'black block' `div`s have a solid black border on one vertical side that is of the required width instead of relying on the background?

jonrandy: Not 100% sure how the 'remove background' feature works, but it's generally to conserve ink/toner by only printing content as opposed to blocks of background colour/image - so I suspect it may well hide css `background`s

includes/gateway/class-omise-payment-billpayment-tesco.php
guzzilar: @jonrandy Come to think of it, what do you think if we put this topic aside for now as it's not serve a majority users?

Supporting plaintext email requires a lot more stuff to do (i.e. generate a new page in WordPress system so we can put the generated page's link to an order confirmation email) and it might not worth to priority it at the moment.

includes/gateway/class-omise-payment-billpayment-tesco.php
guzzilar: @jonrandy  The "border width" approach is pretty interesting.
Let me test a bit more on this.

Thanks for the suggestion 👍

jonrandy: Adding
```
box-sizing: border-box
width: 2px
border-left: 2px solid #000
```
to a `div` gets you a solid div 2px wide without using a background colour

guzzilar: @jonrandy 👍 cool!
I didn't go with `box-sizing`. Last time I checked, many email clients don't support this feature, except Gmail. But done. Can you have a look again?

Thanks :D

guzzilar:
@jonrandy I made a change here https://github.com/omise/omise-woocommerce/pull/125/commits/91b739f31cc4efbf7baedd13989e24e3e3f855bc#diff-b100586f82e3090b17f2369adb8d9235R202

Tested on Gmail, Desktop & iPhone.
Screen Shot 2562-08-08 at 03 47 21

includes/gateway/class-omise-payment-billpayment-tesco.php
```diff
@@ -109,32 +110,109 @@ public function result( $order_id, $order, $charge ) {
...
+
+			// Read data from all <rect> nodes.
+			foreach ( $xml->g->g->children() as $rect ) {
+				$attributes = $rect->attributes();
+				$width      = $attributes['width'];
+				$margin     = ( $attributes['x'] - $prevX - $prevWidth ) . 'px';
+
+				// Set HTML attributes based on <rect> node's attributes.
+				$divRect = $xhtml->createElement( 'div' );
+				$divRect->setAttribute( 'style', "float: left; position: relative; height: 50px; border-left: $width solid #000000; width: 0; margin-left: $margin" );
```
jonrandy: Do we have a potential problem with barcodes not being visible on themes that have a black background? The HTML here does not seem to have a white background behind the bars

guzzilar: @jonrandy if you ask me, super rare. There is a white background at the confirmation email by itself.
Well, I can add a white background here to make sure that it will prevent for those edge cases, probably a new PR after #126 should be ok? (just to avoid merging conflict).

jonrandy: New PR is ok, but should probably be done before release

guzzilar: sure, right after #125 and #126. I'm planning to submit one more for adjusting UI and add some more missing info.

jacstn:
i think email should be in easy to print format, and should not have any other data than Tesco code.

why people from lotus could see my address or what I bought.

PR #126, opened 2019-08-05
 Bill Payment: Adding a 'Print barcode' button.

guzzilar:
> ⚠️ THIS PULL REQUEST REQUIRES #122 AND #125 TO BE MERGED FIRST.

#### 1. Objective

Regarding on user behaviour when it comes to a barcode payment.
Either the barcode has to be mobility-supported (which we already provided a solution by sending barcode through email at #125.), or user should be able to print it and carry out.

This pull request is adding a printing button that would help saving our end-users from remembering and finding a `ctrl+p` shortcut on their keyboard.

#### 2. Description of change

Adding a printing button at WooCommerce Thank You page.

#### 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: v3.6.5
- **WordPress**: v5.2.2
- **PHP**: v7.2.20

**✏️ Details:**

Make a purchase using `Bill Payment: Tesco` payment method as normal.
At the last step, "Thank You" page, there should be a 'printing barcode' button appear under the barcode reference number.
printing-btn

Clicking at the button, a browser printing feature should be triggered, plus, the printing button should be disappeared at the printing page.
printing-btn-2

#### 4. Impact of the change

No

#### 5. Priority of change

Normal

#### 6. Additional Notes

No

### DIFF
--- a/includes/gateway/class-omise-payment-billpayment-tesco.php
+++ b/includes/gateway/class-omise-payment-billpayment-tesco.php
@@ -160,47 +160,50 @@ function register_omise_billpayment_tesco() {
 			<?php if ( 'email' !== $context ) : ?>
 
 				<div class="omise-billpayment-tesco-print-detail">
-					<p class="woocommerce-notice woocommerce-notice--success woocommerce-thankyou-order-received"><?php echo apply_filters( 'woocommerce_thankyou_order_received_text', __( 'Thank you. Your order has been received.', 'woocommerce' ), $order ); ?></p>
-					<ul class="woocommerce-order-overview woocommerce-thankyou-order-details order_details">
+					<div>
+						<h2><?php echo apply_filters( 'woocommerce_thankyou_order_received_text', __( 'Thank you. Your order has been received.', 'woocommerce' ), $order ); ?></h2>
+					</div>
 
-						<li class="woocommerce-order-overview__order order">
+					<div class="omise-billpayment-tesco-print-order-overview">
+						<div class="omise-billpayment-tesco-print-order-overview-item">
 							<?php _e( 'Order number:', 'woocommerce' ); ?>
-							<strong><?php echo $this->order()->get_order_number(); ?></strong>
-						</li>
+							<br/><strong><?php echo $this->order()->get_order_number(); ?></strong>
+						</div>
 
-						<li class="woocommerce-order-overview__date date">
-							<?php _e( 'Date:', 'woocommerce' ); ?>
-							<strong><?php echo wc_format_datetime( $this->order()->get_date_created() ); ?></strong>
-						</li>
+						<div class="omise-billpayment-tesco-print-order-overview-item">
+							<?php _e( 'Date:', 'woocommerce' ); ?>:
+							<br/><strong><?php echo wc_format_datetime( $this->order()->get_date_created() ); ?></strong>
+						</div>
 
 						<?php if ( is_user_logged_in() && $this->order()->get_user_id() === get_current_user_id() && $this->order()->get_billing_email() ) : ?>
-							<li class="woocommerce-order-overview__email email">
-								<?php _e( 'Email:', 'woocommerce' ); ?>
-								<strong><?php echo $this->order()->get_billing_email(); ?></strong>
-							</li>
+							<div class="omise-billpayment-tesco-print-order-overview-item">
+								<?php _e( 'Email:', 'woocommerce' ); ?>:
+								<br/><strong><?php echo $this->order()->get_billing_email(); ?></strong>
+							</div>
 						<?php endif; ?>
 
-						<li class="woocommerce-order-overview__total total">
-							<?php _e( 'Total:', 'woocommerce' ); ?>
-							<strong><?php echo $this->order()->get_formatted_order_total(); ?></strong>
-						</li>
+						<div class="omise-billpayment-tesco-print-order-overview-item">
+							<?php _e( 'Total:', 'woocommerce' ); ?>:
+							<br/><strong><?php echo $this->order()->get_formatted_order_total(); ?></strong>
+						</div>
+					</div>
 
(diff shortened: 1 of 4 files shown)


jacstn:
Works... only one remark you could consider..
why we force user to print 4 pages of text to print one barcode? why he need all that data?
he needs only barcode, name of shop probably.. amount... that is it.

jacstn:
email layout is not good.. as there are elements that are covering text, there is a danger that barcode will be hidden...

image

barcode might get hidden depends on information that is above this barcode.. it might be pushed below this 3 icon bar.

image


This template could be simplified to 1 page, currently, in my wp configuration, there are 4 pages to print.

includes/gateway/class-omise-payment-billpayment-tesco.php
```diff
@@ -0,0 +1,232 @@
...
+						$charge['source']['references']['reference_number_1'],
+						$charge['source']['references']['reference_number_2'],
+						$charge['amount']
+					);
+					?>
+				</small>
+
+				<?php if ( 'email' !== $context ) : ?>
+					<div class="omise-billpayment-tesco-print-button">
+						<button onClick="window.print()" class="button button-primary">Print barcode</button>
```
jonrandy: This will print the entire page - not just the barcode. We should probably consider printing just the barcode ans some relevant info (as @jacstn already mentioned). This could probably be achieved by means of a print stylesheet temporarily applied, or perhaps creating an invisible iframe with some programmatically generated content in it, and printing that

guzzilar: @jacstn @jonrandy thanks for the input.
I intended to leave it as it is because I want to let a stylesheet of a store's theme to handle the printing style in case if they want to have some product references on a paper when it get printed out. (i.e. product detail, price).

Having a simple print function is leaving a room for merchant/developer to remove it out later on if they want to (and user is still able to choose to print only a single page from a browser's printing dialog anyway) – on the other hand, forcing to print only barcode leaving no room for any modification (it could, but adding it back is technically more troublesome than removing it out). 

@jacstn @jonrandy ☝️ with the above reason, what do you guys think, still suggest to force the page to be printed out only barcode?

jacstn: not sure who wants to print anything more than a barcode, to pay to store,
but it is not a big issue, so I am not going to argue about that. AirAsia makes you print the only barcode with brief info what you should do with that.

@guzzilar how about this issue with making a layout that might introduce some problems if not customized by a merchant?

I mean.. this part with some extra icons..

jonrandy: I still think we should really print just the barcode. If we are going to leave as it is now though, I would relabel the button 'Print this page' to give the user some warning they are about to kill more trees than is necessary

jonrandy: If we programmatically create something in a hidden iframe, there is no issue with any conflicts with the merchant's checkout page design... but they would have no control over the way the barcode printout looked. Unless we could somehow pull a themeable layout from the server into said hidden iframe, fill in the details and barcode programmatically, then trigger a print there

guzzilar: @jacstn @jonrandy 👍 thanks, ok, let's try with you guys suggested.
I'll add a hidden `<div></div>` element and set it to only display on `@media print` with some order informations.

After this, I have one PR that is allowing WordPress theme to override Omise template (#116). This should help merchant to be able to custom the printing detail later on.

guzzilar: Ok, I said that let's try but after looked at a printing style of a theme I'm currently using:
Screen Shot 2562-08-08 at 04 41 39

Screen Shot 2562-08-08 at 04 41 48

I think it's really a theme responsibility to proper style the printing page, not from plugin...
How about just change a button title to `print this page`?

👆 if you guys still vote for having a hidden div then I'll give up. But I'll pretty much just put everything that 90% similar to the above screenshots anyway... (payment detail, barcode, order detail).

guzzilar: Screen Shot 2562-08-08 at 04 48 16
btw, here is the actual screen of the above printing screen - screenshots.

guzzilar: Ok, second thought after second thought (should it says "third thought"? lol). 
You guys are right, at least plugin should responsible for providing a default one with a proper way that 3rd-party could override and customize it at their desire.

Thanks for all the inputs 👍 👍

guzzilar:
fyi, @jacstn @jonrandy here is the final printing page for barcode. Code is coming
[billpayment-barcode.pdf](https://github.com/omise/omise-woocommerce/files/3479501/billpayment-barcode.pdf)

guzzilar:
Pushed.

jonrandy:
Doesn't this latest solution have the effect of preventing the customer from being able to print the whole page normally, should they want to do that?

guzzilar:
@jonrandy adding event listener to catch an event after finish printing and/or cancel a printing dialog?

jonrandy:
Not exactly sure how it would work - would need to test. I'm not sure if `window.print()` will print the state of the page when it is called, or the state of the page when printing actually starts. You'd need to make use of the disabled property of the stylesheet too

Unfortunately, Safari is an a**hole - https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onafterprint

guzzilar:
@jonrandy yup, seems it's a bit over design for now as a current stage that we have no concrete feedback just yet. How about we just go with what we have, gather feedback, and come back to improve it if necessary?

jonrandy:
Pretty easy to implement, but I guess no real harm in holding off for now

PR #146, opened 2019-11-06
Credit Card, automatically detect if a particular charge is processing using 3-D Secure feature

guzzilar:
### 1. Objective

A mistake has been happened repeatedly when it comes to enable 3-D Secure feature on Omise plugin. As the current flow of enabling 3-D Secure on Omise-WooCommerce is as the following:
1. Contact Omise support to enable the feature (or Omise support team may be the one who initiate a contact to merchant first, vice versa)
2. Omise enables the 3-D Secure feature, then inform back to merchant.
3. Merchant needs to go back to Omise-WooCommerce setting page, then enable 3-D Secure feature at the Credit Card Payment Setting page.

However, the third step is where thing gets complicated and mistake occurs. Either "enable 3-D Secure on the plugin but not on Omise account" or "enable 3-D Secure on Omise account but not on Omise plugin".

This pull request is to reduce the amount of this issue by programmatically detect if a particular charge is likely to be processed with 3-D Secure feature, then automatically redirect buyers out without having to alter any setting.

**Related information**:
Related issue(s): T17199, T17237 (internal ticket)

### 2. Description of change

1. Removing **"3-D Secure"** option out from the Credit Card payment setting page.
Screen Shot 2562-11-06 at 06 49 41

2. Programmatically detect if a particular charge (using credit card as a payment method) is processing using 3-D Secure feature (or likely to be).

### 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: v3.8.0
- **WordPress**: v5.3.2
- **PHP version**: 7.3.3

**✏️ Details:**

Making sure that there is no side-effect on a charge that does not require 3-D Secure payment.

> Note that Omise-WooCommerce is using Omise API version 2017.

✅ **1. Create a new charge using Omise Japan account with Credit Card payment method.**
**<ins>Expectation</ins>**
A new charge should be created properly the same as before this change, and there must not be any redirection happening during the process.
Screen Shot 2562-11-07 at 04 24 23
Screen Shot 2562-11-07 at 04 24 59
Screen Shot 2562-11-07 at 04 26 15
Screen Shot 2562-11-07 at 04 33 25

✅ **2. Create a new charge using Omise Singapore account with Credit Card payment method.**
**<ins>Expectation</ins>**
A new charge should be created properly the same as before this change, and there must not be any redirection happening during the process (as this test is using an account that 3-D Secure feature isn't enabled).
Screen Shot 2562-11-07 at 04 40 48
Screen Shot 2562-11-07 at 04 41 18
Screen Shot 2562-11-07 at 04 42 04
Screen Shot 2562-11-07 at 04 42 38

✅ **3. Create a new charge using Omise Thailand account "WITHOUT 3-D Secure" with Credit Card payment method.**
**<ins>Expectation</ins>**
A new charge should be created properly the same as before this change, and there must not be any redirection happening during the process.
Screen Shot 2562-11-07 at 04 54 18
Screen Shot 2562-11-07 at 04 55 00
Screen Shot 2562-11-07 at 04 55 27 copy
Screen Shot 2562-11-07 at 04 56 11

✅ **4. Create a new charge using Omise Thailand account "WITH 3-D Secure enabled" with Credit Card payment method.**
**<ins>Expectation</ins>**
A new charge should be created properly the same as before this change, and plugin should be able to detect that a particular charge is in a condition that redirection is needed – user should be redirected out to `authorize_uri`.
Screen Shot 2562-11-07 at 05 01 37
Screen Shot 2562-11-07 at 05 01 50
Screen Shot 2562-11-07 at 05 02 04
Screen Shot 2562-11-07 at 05 02 26 copy
Screen Shot 2562-11-07 at 05 02 39

✅ **5. Searching, if this `omise_3ds` option is being used anywhere else**
**<ins>Expectation</ins>**
Should found nothing. 
Screen Shot 2562-11-07 at 03 22 37
Screen Shot 2562-11-07 at 03 22 33

### 4. Impact of the change

There should be no negative impact. 
However, it would tremendously positively impact with Omise support team and merchants to reduce the communication regarding on "how to turn on 3-D Secure mode in Omise-WooCommerce plugin".

### 5. Priority of change

Normal

### 6. Additional Notes

**! IMPORTANT;** This is to remove the setting out from the user interface. 
However, the option that has already been saved within WordPress DB for current merchants that have been using Omise-WooCommerce plugin, will still be existed.

Screen Shot 2562-11-06 at 07 03 06

Using this query:
`SELECT option_value FROM wp_options WHERE option_name='woocommerce_omise_settings';`
The screenshot above shows that the option `omise_3ds` is existed, its value could be either `yes` or `no`. However, it shall have no effect to this change.

**Why not remove it?**
You may ask. The reason is that I don't want this PR gets too elaborated. Touching on merchant's DB is a bit risky while the actual fix doesn't need to do that. 

We may provide a different PR to clean up Omise settings later on, if needed.
...

Also, note that this would benefit for the dynamic 3-D Secure as well.

### DIFF
--- a/includes/gateway/class-omise-payment-creditcard.php
+++ b/includes/gateway/class-omise-payment-creditcard.php
@@ -35,7 +35,6 @@ function register_omise_creditcard() {
 
 			$this->title          = $this->get_option( 'title' );
 			$this->description    = $this->get_option( 'description' );
-			$this->omise_3ds      = $this->get_option( 'omise_3ds', false ) == 'yes';
 			$this->payment_action = $this->get_option( 'payment_action' );
 
 			add_action( 'woocommerce_api_' . $this->id . '_callback', array( $this, 'callback' ) );
@@ -91,16 +90,6 @@ function register_omise_creditcard() {
 							self::PAYMENT_ACTION_AUTHORIZE         => __( 'Manual Capture', 'omise' )
 						),
 						'desc_tip'    => true
-					),
-					'omise_3ds' => array(
-						'title'       => __( '3-D Secure payment', 'omise' ),
-						'type'        => 'checkbox',
-						'label'       => __( 'Enabling 3-D Secure payment means that buyer will be redirected to the 3-D Secure authorization page during the checkout process.', 'omise' ),
-						'description' => wp_kses(
-							__( 'Please be informed that you must contact Omise support team (support@omise.co) before enable or disable this option.<br/> (Japan-based accounts are not eligible for the service.)', 'omise' ),
-							array( 'br' => array() )
-						),
-						'default'     => 'no'
 					),
 					'accept_visa' => array(
 						'title'       => __( 'Supported card icons', 'omise' ),
@@ -291,7 +280,8 @@ function register_omise_creditcard() {
 				return $this->payment_failed( Omise_Charge::get_error_message( $charge ) );
 			}
 
-			if ( $this->omise_3ds ) {
+			// If 3-D Secure feature is enabled, redirecting user out to a 3rd-party credit card authorization page.
+			if ( self::STATUS_PENDING === $charge['status'] && ! $charge['authorized'] && ! $charge['paid'] && ! empty( $charge['authorize_uri'] ) ) {
 				$order->add_order_note(
 					sprintf(
 						__( 'Omise: Processing a 3-D Secure payment, redirecting buyer to %s', 'omise' ),


jonrandy:
It's actually possible to simplify this even further. Just do the redirect every time. Don't check anything

guzzilar:
I was considering about that approach. But kinda still want to achieve our main key feature, the white-label one.

jonrandy:
I don't know what you mean

guzzilar:
I mean, one of the Omise key features is to have an experience of not redirect buyers out from a site. I wanted to achieve that goal by not redirect buyers out of a store if it does not need to.

Screen Shot 2562-11-07 at 10 37 33

jacstn:
tested, works good..

jonrandy:
Fair enough, but the redirect is for a split second and there is no impression they ever left the site - especially if implemented in the way that Visa and the EU advise - in an iframe. It just seems like an unnecessary bit of code to achieve something extremely minor. Works though, if unduly complex. Approving

guzzilar:
>  but the redirect is for a split second (certainly not distracting) and there is no impression they ever left the site

I think this is working on Shopify because its payment behaviour. As it's kinda default thing to redirect to a payment-provider checkout form in Shopify, so we have a perception of redirecting between site to site already – adding one more loading screen or a split second seems to be nothing.

However, thanks for reviewing ;)

PR #151, opened 2020-01-19
Payment Setting: properly display payment methods based on a given Omise Account (for admin)

guzzilar:
### 1. Objective

Previously all the payment methods that have been implemented to the plugin will be shown at WooCommerce Payment Setting page, regardless on a given Omise account.

For instance, using Omise Japan account at Omise-WooCommerce plugin, you will see all the payment methods, including those that can be used only for Omise Thailand.

This, is believed to cause a bad user-experience and a confusion for all the merchants outside Thailand.

**Related information**:
Related issue(s): T18705

### 2. Description of change

1. Before saving given credential keys to WooCommerce's DB, the plugin will be making a request to Omise API to retrieve Omise Account detail, then pack and save it to WooCommerce's DB along with those credentials.

At WordPress's `wp_options` table, option_name: `woocommerce_omise_settings`, there will be 3 new parameters: `account_id`, `account_email`, and `account_country`.

**Before**
```
a:5:{s:7:"sandbox";s:3:"yes";s:15:"test_public_key";s:29:"pkey_test_***";s:16:"test_private_key";s:29:"skey_test_***";s:15:"live_public_key";s:0:"";s:16:"live_private_key";s:0:"";}
```

**After**
```
a:8:{s:10:"account_id";s:24:"acct_***";s:13:"account_email";s:22:"***@***";s:15:"account_country";s:2:"JP";s:7:"sandbox";s:3:"yes";s:15:"test_public_key";s:29:"pkey_test_***";s:16:"test_private_key";s:29:"skey_test_***";s:15:"live_public_key";s:0:"";s:16:"live_private_key";s:0:"";}
```

.
2. At the Admin Payment Setting page, add a condition to display only payment methods that are available for that specific Omise account's country.

**Before**
_The below screenshot: all the payment methods are listed at the payment setting page regardless on a given Omise account._
Screen Shot 2563-01-20 at 06 35 45

**After**
_The below screenshot: hide all the payment methods if Omise account's credential hasn't been set._
Screen Shot 2563-01-20 at 06 34 32

_The below screenshot: Once the Omise account has been set, the payment methods will be shown accordingly to the given account (in this case, Thailand account)._
Screen Shot 2563-01-20 at 06 39 55

_The below screenshot: Once the Omise account has been set, the payment methods will be shown accordingly to the given account (in this case, Japan account)._
Screen Shot 2563-01-20 at 06 40 45

3. Also make sure that it won't affect to the current merchants.

#### 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: v3.8.1
- **WordPress**: v5.3.2
- **PHP version**: 7.3.3

**✏️ Details:**

There are 2 possible cases.
1. Current users upgrading from the previous plugin version to this one.
The following code should detect and auto-update its database to meet with the condition of displaying payment methods by a given Omise account.
https://github.com/omise/omise-woocommerce/pull/151/files#diff-9849043e36d1d7b568f325534b722341R77

As for user perspective, they wouldn't need to do any extra step.

2. New users that download the plugin for the first time.
This group should be no problem as described at the section **2. Description of change**, [2].

#### 4. Impact of the change

As described at the section **2. Description of change**, [1]. There will be 3 new parameters saved at WordPress's `wp_options` table, option_name: `woocommerce_omise_settings`.

#### 5. Priority of change

Normal

#### 6. Additional Notes

No

### DIFF
--- a/includes/admin/views/omise-page-settings.php
+++ b/includes/admin/views/omise-page-settings.php
@@ -98,85 +98,89 @@
 		<hr />
 
 		<h3><?php _e( 'Payment Methods', 'omise' ); ?></h3>
-		<p><?php _e( 'The table below is a list of available payment methods that you can enable in your WooCommerce store.', 'omise' ); ?></p>
-		<table class="form-table">
-			<tbody>
-				<tr>
-					<th scope="row"><label for="sandbox"><?php _e( 'Available Payment Methods', 'omise' ); ?></label></th>
-					<td>
-						<table class="widefat fixed striped" cellspacing="0">
-							<thead>
-								<tr>
-									<?php
-										$columns = array(
-											'name'    => __( 'Payment Method', 'omise' ),
-											'status'  => __( 'Enabled', 'omise' ),
-											'setting' => ''
-										);
+		<?php if ($settings['account_country']) : ?>
+			<p><?php _e( 'The table below is a list of available payment methods that you can enable in your WooCommerce store.', 'omise' ); ?></p>
+			<table class="form-table">
+				<tbody>
+					<tr>
+						<th scope="row"><label for="sandbox"><?php _e( 'Available Payment Methods', 'omise' ); ?></label></th>
+						<td>
+							<table class="widefat fixed striped" cellspacing="0">
+								<thead>
+									<tr>
+										<?php
+											$columns = array(
+												'name'    => __( 'Payment Method', 'omise' ),
+												'status'  => __( 'Enabled', 'omise' ),
+												'setting' => ''
+											);
 
-										foreach ( $columns as $key => $column ) {
+											foreach ( $columns as $key => $column ) {
+												switch ( $key ) {
+													case 'status' :
+													case 'setting' :
+														echo '<th style="text-align: center; padding: 10px;" class="' . esc_attr( $key ) . '">' . esc_html( $column ) . '</th>';
+														break;
+
+													default:
+														echo '<th style="padding: 10px;" class="' . esc_attr( $key ) . '">' . esc_html( $column ) . '</th>';
+														break;
+												}
+
+											}
+										?>
+									</tr>
+								</thead>
+								<tbody>
+									<?php
+									$available_gateways = array(
+										new Omise_Payment_Alipay,
+										new Omise_Payment_Billpayment_Tesco,
+										new Omise_Payment_Creditcard,
+										new Omise_Payment_Installment,
+										new Omise_Payment_Internetbanking,
+										new Omise_Payment_Truemoney
+									);
+									foreach ( $available_gateways as $gateway ) :
+
+										echo '<tr>';
+
+										foreach ( $columns as $key => $column ) :
 											switch ( $key ) {
-												case 'status' :
-												case 'setting' :
-													echo '<th style="text-align: center; padding: 10px;" class="' . esc_attr( $key ) . '">' . esc_html( $column ) . '</th>';
+												case 'name' :
+													$method_title = $gateway->get_title() ? $gateway->get_title() : __( '(no title)', 'omise' );
+													echo '<td class="name">
+														<a href="' . admin_url( 'admin.php?page=wc-settings&tab=checkout&section=' . strtolower( $gateway->id ) ) . '">' . esc_html( $method_title ) . '</a>
+													</td>';
 													break;
 
-												default:
-													echo '<th style="padding: 10px;" class="' . esc_attr( $key ) . '">' . esc_html( $column ) . '</th>';
+												case 'status' :
+													echo '<td class="status" style="text-align: center;">';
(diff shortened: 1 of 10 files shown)


includes/admin/views/omise-page-settings.php
```diff
@@ -98,85 +98,91 @@
...
-						<table class="widefat fixed striped" cellspacing="0">
-							<thead>
-								<tr>
-									<?php
-										$columns = array(
-											'name'    => __( 'Payment Method', 'omise' ),
-											'status'  => __( 'Enabled', 'omise' ),
-											'setting' => ''
-										);
+		<?php if ($settings['account_country']) : ?>
```
guzzilar: Note; this seems to be a lot of changes but it only is because of this line and another following line.
The rest is about the changing of its indent.

includes/admin/views/omise-page-settings.php
```diff
@@ -98,85 +98,91 @@
...
+									$available_gateways = array(
+										new Omise_Payment_Alipay,
+										new Omise_Payment_Billpayment_Tesco,
+										new Omise_Payment_Creditcard,
+										new Omise_Payment_Installment,
+										new Omise_Payment_Internetbanking,
+										new Omise_Payment_Truemoney
+									);
+									foreach ( $available_gateways as $gateway ) :
+										if ( $gateway->is_country_support( $settings['account_country'] ) ) :
```
guzzilar: Note; continue from https://github.com/omise/omise-woocommerce/pull/151/files#r368338668
Here is another line that has been added. The rest is code indentation.

jonrandy:
Would this not be better achieved using the Capability API? This would make it more future proof by preventing the need to change the code should the country availability of a payment method change.

Admittedly, the downside to this is that some payment methods *still* do not appear in the capability API, so we'd need to wait for that to be done before release (but I believe it should be done soon).

Arguably, the best solution would be a combination of using the Capability API, and what you have done. We could show all the payment methods available for the merchant country overall, and grey out the ones that are not currently available to that particular merchant

guzzilar:
@jonrandy Thanks for the input 👍
The use case of this approach is slightly different from using Capability API.

The Capability API is to list out available payment methods of that particular Omise account "at the moment". But what this pull request is trying to achieve here is to list all the payment methods that is available for that "country" regardless of if it can be used on that particular account or not.

However, I would love to implement the Capability API once it done, to maximize this feature so the output can be:
```
TH Account:
- Available payment methods
  - Alipay
  - Bill Payment [disabled] (please contact Omise support to enable this payment)
  - Credit Card
  - Internet Banking
  - Installment
  - TrueMoney Wallet

JP Account:  
- Available payment methods
  - Credit Card
  - Konbini [disabled] (please contact Omise support to enable this payment)
```

"Here are 5 payment methods that you may use with this account, currently 3 enabled and 2 disabled. To enable these 2 payment methods, please contact our support team".

jonrandy:
Yup, that was my arguable 'best solution' above that is the best of both worlds

guzzilar:
@jonrandy ah sorry I miss-read the whole sentence of `and grey out the ones that are not currently available to that particular merchant`. You've already mentioned what I've explained in a shorter and better sentence. Yeah, exactly like that, in my opinion.

includes/admin/views/omise-page-settings.php
```diff
@@ -98,85 +98,91 @@
...
+								</thead>
+								<tbody>
+									<?php
+									$available_gateways = array(
+										new Omise_Payment_Alipay,
+										new Omise_Payment_Billpayment_Tesco,
+										new Omise_Payment_Creditcard,
+										new Omise_Payment_Installment,
+										new Omise_Payment_Internetbanking,
+										new Omise_Payment_Truemoney
```
jonrandy: Wow - they're even alphabetised! :)

jonrandy: I do have the feeling this list of available gateways really belongs at a higher level as it feels like something that may be used again elsewhere. It's ok here for now though I guess

guzzilar: Thanks for the input 👍 
Now you have mentioned and it got me to think that probably the main `Omise` class (omise-woocommerce.php) would also likely to be a place where this can also be implemented if not too complex.

Note for idea (but not in anytime soon)
```php
foreach ( Omise::available_payment_methods() as $methods ) {
    // ...
}
```

guzzilar: also note for myself. I think `gateways` is a bad name. Should be `methods`

jonrandy: Yep. I mentioned it because I'm doing something similar in my Prestashop refactoring

jacstn:
works great..

guzzilar:
@jacstn @jonrandy Thanks for the review! 👍

PR #149, opened 2019-11-27
Introducing Konbini payment method

guzzilar:
## 1. Objective

Omise has been providing a payment service (credit card) for Japan for a couple years. And now, we have also delivered a brand-new payment method, "Konbini" recently that would enable more opportunities for merchants and buyers that are living in Japan to have more choice of payments and to make the payment experience even better.

Now it's time to bring this new payment feature to Omise-WooCommerce plugin

**Related information**:
Related issue(s): T17761 (internal ticket)

## 2. Description of change

- Displaying **"Konbini"** payment method on both Admin and checkout page.
- Merchant will be able to enable **"Konbini"** payment start accepting payment through Konbini.
- Buyers will be able to select **"Konbini"** as a payment method when checking out an order. 

## 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: v4.3.3
- **WordPress**: v5.4.2
- **PHP version**: 7.3.3

**✏️ Details:**

Screen Shot 2563-08-19 at 00 23 58

Screen Shot 2563-08-19 at 00 25 26

Screen Shot 2563-08-19 at 00 41 22

## 4. Impact of the change

none

## 5. Priority of change

Normal

## 6. Additional Notes

As this payment method is aiming for Japanese market, this plugin should be translated in Japanese language before release.

PR #152, opened 2020-02-25
Introducing PayNow payment method (only available in Singapore)

guzzilar:
### 1. Objective

As Omise has been providing a support for PayNow payment method.
This pull request is to integrate the new PayNow payment to WooCommerce plugin.

**Related information**:
Related issue(s): T19391 (internal ticket)

### 2. Description of change

> **IMPORTANT**: The changes only appear when using Omise Singapore account.

1. Integrating Omise PayNow payment method with WooCommerce.
At Omise Settings page, you will see a new payment method, **"PayNow".**
Screen Shot 2563-02-26 at 12 04 44 copy

Screen Shot 2563-02-24 at 09 50 20

2. Once enabled the payment, the PayNow payment method will be appeared at WooCommerce Checkout page.
Screen Shot 2563-02-25 at 16 48 09

3. Displaying PayNow QR code at the order-confirmation page, once placing an order with PayNow payment method.
Screen Shot 2563-02-25 at 16 48 37

4. With **Omise Webhook** feature, once the charge has been paid, an order status will be updated to `processing`.
Screen Shot 2563-02-25 at 17 00 23

### 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: -
- **WordPress**: v5.3.2
- **PHP version**: -

**✏️ Details:**

> **IMPORTANT**: To test this feature, requires Omise Singapore account, and also the Webhook feature should be set on that particular account.

**1. Making sure that the payment method's setting is available only on Omise Singapore account.**

**Using Omise Singapore account**
Screen Shot 2563-02-26 at 12 04 44 copy

**Using Omise Japan account**
Screen Shot 2563-02-26 at 12 03 40

**2. Making sure that once enabled the PayNow payment method, the method will be displayed at WooCommerce Checkout page properly.**
Screen Shot 2563-02-25 at 16 48 09

**3. Making sure that once the order has been placed, a PayNow QR code will be displayed at the Order Received page, scannable.**
Screen Shot 2563-02-25 at 16 48 37 copy

The example using `QR Reader` application, on iPhone
Screen Shot 2563-02-26 at 12 15 45

An order has been placed with `on-hold` order status.
Screen Shot 2563-02-26 at 12 18 47 copy

**4. Making sure that Omise `charge.complete` event will be caught properly once the payment has been paid.**

To simulate this situation, firstly Omise Webhook feature must be promptly set with a store's webhook endpoint.
Screen Shot 2563-02-26 at 12 23 53 copy

Then to go to Omise Dashboard and mark a particular charge as successful.
Screen Shot 2563-02-26 at 12 22 49 copy

Once done, the plugin will be able to catch a `charge.complete` event that has been triggered by Omise, then update an order status accordingly.
Screen Shot 2563-02-26 at 12 26 34 copy

_The following graphic is showing an overview of the payment flow._
paynow-flow

### 4. Impact of the change

None

### 5. Priority of change

Normal

### 6. Additional Notes

This pull request is providing the least basic feature of PayNow payment method.
There are still numbers of thing that can be added to make its experience better however will be discussed and implemented later. 

i.e. 
1. An input from Irista suggested a counting down mechanism to automatically cancel the payment & order after a certain time if the payment hasn't been paid (10-15 minutes). 

2. Adding more text or footnote to provide an instruction of how to use PayNow.

3. Including QR code to the WooCommerce Order Confirmation email.

and so on.


includes/gateway/class-omise-payment-paynow.php
```diff
@@ -0,0 +1,153 @@
+<?php
```
jonrandy: Looking at a lot of these `class-omise-payment-xxxxx.php` files, there is a lot of replication. Should we consider a hierarchy of classes here - to prevent further duplication and simplify existing classes?

omise-woocommerce.php
```diff
@@ -81,6 +81,7 @@ public function init() {
 		register_omise_creditcard();
```
jonrandy: Feels like we should have a list of payment methods, then iterate over a list to call these `register_omise_xxxx` functions. Such a list would also be beneficial for including the files etc. - when a new payment method was added, you could simply add a new name to the list

includes/admin/views/omise-page-settings.php
```diff
@@ -139,6 +139,7 @@
 										new Omise_Payment_Creditcard,
 										new Omise_Payment_Installment,
 										new Omise_Payment_Internetbanking,
```
jonrandy: Here again, feels like this code could be simplified by having a general list of available payment methods somewhere. Then you could just update the list instead of updating this code

guzzilar: @jonrandy how do you feel about this approach? 
https://github.com/omise/omise-woocommerce/pull/153/files

includes/gateway/class-omise-payment-paynow.php
guzzilar: I could also see some part of the code that could be lean a bit more (`__construct`, `charge`, and `result` methods). But as for business perspective, probably not so efficient if we to do it at this time (as refactoring will take a lot of time to discuss and review).

But noted and acknowledged that we could improve those `class-omise-payment-xxxxx.php` classes a bit more.
I'll look into it after #149, #152 and one more coming PR, "API localization" get merged and released.
Thanks 👍

PR #153, opened 2020-02-27
(proposal) Code cleaning for payment method classes.

guzzilar:
### 1. Objective
This code is a proposal of removing an unnecessary code for registering and listing up Omise payment methods.

**Related information**:
Related issue(s): T19568 (internal ticket)

### 2. Description of change

Refactoring part of code that was used for registering Omise payment methods to WooCommerce.

As well as, removing the redundant code.

### 3. Quality assurance

1. Making sure that all the payment methods can be listed out at the WooCommerce payment setting page.

2. Making sure that all the payment methods can be used as normal at WooCommerce checkout page.

### 4. Impact of the change

None

### 5. Priority of change

Normal

### 6. Additional Notes

### DIFF
--- a/includes/admin/views/omise-page-settings.php
+++ b/includes/admin/views/omise-page-settings.php
@@ -133,15 +133,8 @@
 								</thead>
 								<tbody>
 									<?php
-									$available_gateways = array(
-										new Omise_Payment_Alipay,
-										new Omise_Payment_Billpayment_Tesco,
-										new Omise_Payment_Creditcard,
-										new Omise_Payment_Installment,
-										new Omise_Payment_Internetbanking,
-										new Omise_Payment_Truemoney
-									);
-									foreach ( $available_gateways as $gateway ) :
+									foreach ( Omise()->payment_methods() as $gateway ) :
+										$gateway = new $gateway;
 										if ( $gateway->is_country_support( $settings['account_country'] ) ) :
 
 											echo '<tr>';
--- a/includes/admin/views/omise-page-settings.php
+++ b/includes/admin/views/omise-page-settings.php
@@ -133,7 +133,7 @@
 								</thead>
 								<tbody>
 									<?php
-									foreach ( Omise()->payment_methods() as $gateway ) :
+									foreach ( Omise()->payment_methods as $gateway ) :
 										$gateway = new $gateway;
 										if ( $gateway->is_country_support( $settings['account_country'] ) ) :
 
--- a/omise-woocommerce.php
+++ b/omise-woocommerce.php
@@ -75,13 +75,8 @@ class Omise {
 		$this->register_post_types();
 		$this->init_admin();
 		$this->init_route();
+		$this->register_payment_methods();
 
-		register_omise_alipay();
-		register_omise_billpayment_tesco();
-		register_omise_creditcard();
-		register_omise_installment();
-		register_omise_internetbanking();
-		register_omise_truemoney();
 		prepare_omise_myaccount_panel();
 	}
 
@@ -96,6 +91,20 @@ class Omise {
 			<p><?php echo __( 'Omise WooCommerce plugin requires <strong>WooCommerce</strong> to be activated.', 'omise' ); ?></p>
 		</div>
 		<?php
+	}
+
+	/**
+	 * @since 3.11
+	 */
+	public function payment_methods() {
+		return array(
+			'Omise_Payment_Alipay',
+			'Omise_Payment_Billpayment_Tesco',
+			'Omise_Payment_Creditcard',
+			'Omise_Payment_Installment',
+			'Omise_Payment_Internetbanking',
+			'Omise_Payment_Truemoney'
+		);
 	}
 
 	/**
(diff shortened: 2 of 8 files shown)


includes/class-omise-payment-methods.php
```diff
@@ -0,0 +1,41 @@
+<?php
```
jonrandy: Does this need to be a singleton? It has no state, it is fully self contained. A very simple class with a public static array of the methods would be much simpler. This is overcomplicating things

omise-woocommerce.php
```diff
@@ -126,17 +128,20 @@ private function include_classes() {
 		require_once OMISE_WOOCOMMERCE_PLUGIN_PATH . '/includes/events/class-omise-event-charge-capture.php';
 		require_once OMISE_WOOCOMMERCE_PLUGIN_PATH . '/includes/events/class-omise-event-charge-complete.php';
 		require_once OMISE_WOOCOMMERCE_PLUGIN_PATH . '/includes/events/class-omise-event-charge-create.php';
+		require_once OMISE_WOOCOMMERCE_PLUGIN_PATH . '/includes/gateway/abstract-omise-payment-offsite.php';
 		require_once OMISE_WOOCOMMERCE_PLUGIN_PATH . '/includes/gateway/class-omise-payment-alipay.php';
```
jonrandy: We now have a list of payment methods - why not use that to automate the `require`s?

includes/class-omise-payment-methods.php
guzzilar: yeah, no particular reason to be singleton. Let me think if this will be in some other uses. If no then will go with what you suggested.

omise-woocommerce.php
guzzilar: Just don't see much benefit on doing that for now.

guzzilar:
@jonrandy I removed the whole 'payment methods' class. Can you help check it again?
Thanks :D

omise-woocommerce.php
jonrandy: Just thinking about not repeating ourselves by effectively writing another list of all the payment methods. Better to have one list so that we just need change that instead of writing more boilerplate code every time a new payment method is added

jonrandy:
I think because you 'force pushed' I'm having trouble working out what you've actually removed as those commits are gone. I always find it best never to force push anything - unless it's a last resort to fix some Git weirdness

guzzilar:
@jonrandy about force push, you're right. I forgot.
The change I've made is to remove the `includes/class-omise-payment-methods.php` (the one that you mentioned about singleton) and having a list of payment method classname at `omise-woocommerce.php` file instead (here: https://github.com/omise/omise-woocommerce/pull/153/files#diff-da400947a51dd0d488fcdb5ac3a4edd4R99).

omise-woocommerce.php
guzzilar: As a context of concern, I don't think there is any better place than this place that should be concerning about what file that needed to be included in the plugin. Unless we implement something similar to `autoload` class, having multiple places to list out files that needed to be included is a bit too much of complexity compare to what it will be achieved.

At least, for now.

jonrandy: I wasn't suggesting requiring the files elsewhere - it can still be done here. I just thought it would  be beneficial to use the list of payment methods we already have to do the `require`s here - i.e. loop over the names, transform the name to the file name, require the file

guzzilar: okay I see. I understand all you suggested but I just don't see that it will greatly benefit more than the current code considering that adding that loop of code here in this place as well.
I've really considered it. Until the point of time where this suggest is shining out and in needed, can we skip this point? 
(although the PayNow PR will be touching this topic again, I still don't see it is in needed to be implemented)

includes/admin/views/omise-page-settings.php
```diff
@@ -133,15 +133,8 @@
...
-										new Omise_Payment_Billpayment_Tesco,
-										new Omise_Payment_Creditcard,
-										new Omise_Payment_Installment,
-										new Omise_Payment_Internetbanking,
-										new Omise_Payment_Truemoney
-									);
-									foreach ( $available_gateways as $gateway ) :
+									foreach ( Omise()->payment_methods() as $gateway ) :
+										$gateway = new $gateway;
 										if ( $gateway->is_country_support( $settings['account_country'] ) ) :
```
jonrandy: This name should be `is_country_supported`

omise-woocommerce.php
```diff
@@ -98,6 +93,20 @@ public function init_error_messages(){
 		<?php
 	}
 
+	/**
+	 * @since 3.11
+	 */
+	public function payment_methods() {
```
jonrandy: Would this be simpler implemented as a public variable (or constant - depending on the PHP version)?

includes/admin/views/omise-page-settings.php
guzzilar: Noted. Thanks for pointing out 👍  
I'll make a change in the later code refactoring PR (as it is not in this scope of change).

omise-woocommerce.php
guzzilar: In the project we can call a function `Omise()` to retrieve an instance of Omise object as in the code `Omise()-> payment_methods()`. 👈  I think this one is simpler already. What do you think?

jonrandy: I think `Omise()->payment_methods` is simpler. I don't see a need for it to be a function

includes/admin/views/omise-page-settings.php
jonrandy: Similarly, I think I saw `restricted_countries` which is storing a list of country codes that the method is restricted **to**. Those countries then aren't restricted - quite the opposite, they are allowed. This should probably be renamed `restricted_to_countries`

guzzilar: IIRC I think I copied the name from somewhere in our plugin projects.
If adding `_to_` makes more sense then maybe we should apply it to every where we are using this as well. 
But noted. Thanks 👍

omise-woocommerce.php
guzzilar: I see. I just pushed the update. Can you review again?

omise-woocommerce.php
```diff
@@ -20,6 +20,20 @@ class Omise {
...
+	 */
+	public $payment_methods = array(
+		'Omise_Payment_Alipay',
+		'Omise_Payment_Billpayment_Tesco',
+		'Omise_Payment_Creditcard',
+		'Omise_Payment_Installment',
+		'Omise_Payment_Internetbanking',
+		'Omise_Payment_Truemoney'
+	);
+
```
guzzilar: Note, I prefer this way instead of constant because it gives a flexibility on altering its payment's classname in the future if we make it extensible. As it's quite common in WordPress world. (i.e. WordPress style is kinda heavily using `apply_filter( blah )` to open for a 3rd-party to make a change as they desire without editing the core code).

PR #157, opened 2020-03-19
Code refactoring, simplifying the 'process_refund' method

guzzilar:
### 1. Objective

The main concept of this refactoring is about implementing the refund feature for Alipay and Installment. However, looking at the current code and found that the `process_refund` method has been implemented separately in Credit Card and TrueMoney Wallet payment methods.

Therefore this pull request is submitted to reduce the duplicated code so we can easily implement the refund feature to those 2 payment methods without producing any more duplicated.

**Related information**:
Related issue(s): T19965 (internal ticket)

### 2. Description of change

- Removing `process_refund` method out from `Omise_Payment_Truemoney` and `Omise_Payment_Creditcard` classes (move to its parent class, **Omise_Payment**).

- Implementing the `process_refund` method back to **Omise_Payment** class (so it can be shared between each payment method class that are supported for the refund feature.

- Tweaking messages that are related to the refund feature. Making it more lean, and less redundant, clearer message.  

### 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: v3.7.1
- **WordPress**: v5.3.2
- **PHP version**: 7.0.16
- **Omise plugin version**: Omise-WooCommerce 3.11 (optional, in case of submitting a new issue)

**✏️ Details:**

To make sure that this refactoring will not break any current feature, there are 3 main points we should cover.
**1. Making sure we can make a refund using TrueMoney Wallet payment method.**
1.1. After placed an order, at WooCommerce Order page, you can create a refund here.
truemoney-refund-step-01

1.2. If succeed, there will be a note that the order has been refunded.
truemoney-refund-step-02

1.3. Double check at Omise Dashboard to make sure that the refund has been done properly.
truemoney-refund-step-03

**2. Making sure we can make a refund using Credit Card payment method.**

card-refund-step1

card-refund-step2

card-refund-step3

**3. Making sure we can make a refund using Credit Card payment method and the plugin can detect and add a proper note accordingly if that refund is being treated as voided.**

card-void-step1

card-void-step2

card-void-step3

Also, testing and making sure that the plugin can handle failure cases properly.

**4. Making sure that the plugin will be raising a proper failure message in case if it cannot retrieve an order object from a given order ID.**
> Note, this case is an extremely rare case, as the only chance that this case would happen is when the database gets corrupted.

> Also note, to test this case, you will need to modify code, at file: 
> **includes/gateway/class-omise-payment.php**
> Line: 240. Add a random string after an order id.

Screen Shot 2563-03-19 at 23 51 15

By doing that, when you try to create a refund, the plugin will raise an error that, it cannot retrieve order from the given order ID.

refund-failed-1

**5. You mat try modify another part of the `process_refund` code to see different cases of failure message. In this case, when the Omise Charge object cannot be retrieve.**
At the same file as the case [4], but adding a random string at the line _250_.
Screen Shot 2563-03-19 at 23 59 51

You will get an error message saying that charge cannot be retrieve.
refund-failed-2

**6. Or, at the line _252_ of the same file, you may add a random amount that make the refund failed.**
Screen Shot 2563-03-20 at 00 01 49

By trying to create a refund with amount that is higher than its charge's amount, you will receive a message saying that refund failed.
refund-failed-3

**7. Also, making sure that those orders that are placed using payment methods that don't support for the refund feature, won't be able to create a refund via Omise.**
Placing an order using Internet Banking payment method will not show "Refund via Omise" button at the order detail page.
refund-no-support

### 4. Impact of the change

Nothing

### 5. Priority of change

Normal

### 6. Additional Notes

This PR is focusing on relocating the `process_refund` method.
There will be 3 more PRs coming.
1. Make use of `$reason` variable that is in the 3rd argument of `process_refund`. (in WooCommerce, you can add a reason of refunding, which we can pass that `reason` to the `metadata` parameter when creating a new Refund object)

2. Adding refund feature to `Alipay` and `Installment` payment methods.

3. As from the internal ticket that WooCommerce cannot sync up refund status from Omise Dashboard. One more dedicated PR will be coming to handle this issue.


guzzilar:
Here is a use case after this refactoring.
https://github.com/omise/omise-woocommerce/pull/158/files#diff-3f1dd1d765734b8bbc1ab3abf4d36ff6R12

includes/gateway/class-omise-payment.php
```diff
@@ -226,6 +226,62 @@ public function process_payment( $order_id ) {
...
+				'omise'
+			);
+
+			return new WP_Error( 'error', sprintf( wp_kses( $message, array( 'br' => array() ) ), $order_id ) );
+		}
+
+		try {
+			$charge = OmiseCharge::retrieve( $order->get_transaction_id() );
+			$refund = $charge->refunds()->create( array(
+				'amount' => Omise_Money::to_subunit( $amount, $order->get_order_currency() )
```
jonrandy: `to_subunits`?

includes/gateway/class-omise-payment.php
```diff
@@ -226,6 +226,62 @@ public function process_payment( $order_id ) {
...
+	 * @param  string $reason
+	 *
+	 * @return boolean True|False based on success, or a WP_Error object.
+	 *
+	 * @see    WC_Payment_Gateway::process_refund( $order_id, $amount = null, $reason = '' )
+	 */
+	public function process_refund( $order_id, $amount = null, $reason = '' ) {
+		if ( ! $order = wc_get_order( $order_id ) ) {
+			$message = __(
+				'Refund failed. Cannot retrieve an order from the given ID: %s. Please try again or you may do a manual refund.',
```
jonrandy: "... with the given ID: ..."

jonrandy: "...or do a manual refund."

guzzilar: Thanks 👍 
Wonder what's the difference between `or you may do a manual refund` and `or do a manual refund`. Just a nuanced or it gives a totally different impression?

includes/gateway/class-omise-payment.php
guzzilar: Noted. This one requires some change that does not related to this PR.
Probably another PR will be better if we want to change this. (but very low priority)

includes/gateway/class-omise-payment.php
jonrandy: "Please try again or you may do a manual refund" is kinda weird sounding. It also has a subtly different meaning overall, which I'm struggling to articulate - something like "if you don't try again, you may inadvertently do a manual refund" - which isn't the intended message

includes/gateway/class-omise-payment.php
```diff
@@ -226,6 +226,62 @@ public function process_payment( $order_id ) {
...
+		try {
+			$charge = OmiseCharge::retrieve( $order->get_transaction_id() );
+			$refund = $charge->refunds()->create( array(
+				'amount' => Omise_Money::to_subunit( $amount, $order->get_order_currency() )
+			) );
+
+			if ( $refund['voided'] ) {
+				$message = sprintf(
+					wp_kses(
+						__( 'Omise: Voided an amount %1$s %2$s.<br/>Refund id is %3$s', 'omise' ),
```
jonrandy, line 258: Should be "...an amount of..." (same applies below)

guzzilar:
@jonrandy thanks for pointing out 👍
Fixed here https://github.com/omise/omise-woocommerce/pull/157/commits/287dcb6e629deda65d0b38e4c4935ede0c198369

PR #167, opened 2020-05-08
Code Refactoring, simplifying Callback function.

guzzilar:
## 1. Objective

Credit Card payment method's callback function has been implemented long time ago since before Omise provided other offsite payment methods.

However, as the payment feature grows, this function remains untouched, duplicated, and complicated.

This pull request is to simplify the function, combining between Offsite callback and Credit Card callback. Yet, also decouple the duplicated code to make the code simpler to integrate with API Localization feature.

**Related information**:
Related issue(s): T20306 (internal ticket)

## 2. Description of change

After listing out 2 callback methods from Credit Card and Offsite payment methods.
There are some part of the code that is sharing the same behaviour, and some that could be merged.
Untitled Diagram (8)

Here is a new structure after merged.
Untitled Diagram (9)

## 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: v4.0.1
- **WordPress**: v5.4.1
- **PHP version**: 7.3.3.

**✏️ Details:**

> Note 1: To test all the callback functions on your local machine, you will need [`ngrok`](https://ngrok.com) in order to provide a public url with HTTPS.

> Note 2: Also, there are 7 payment methods that we should make sure that the change on Callback won't effect to its functionality.

> Note 3: We will be testing 3 cases in each payment method, which are, **'successful charge'**, **'failed charge'**, **'pending charge'**.

> Note 4: To not make this article too long, it will skip to explain all the prerequisite steps and focus on the output of the Callback function.

The test can be spilt into 3 main categories, the payment methods that are in the same category will behave the same in its callback function.

### 3.1. Offsite
Those payment methods that require buyer to be redirected out to the 3rd-party payment page.
Payment method: `Alipay`, `Installment`, `Internet Banking`, `TrueMoney Wallet`.

3.1.1). ✅ **Successful** case
> `charge.status = successful`, `charge.authorized = true`, `charge.paid = true`
alipay-successful

3.1.2). ✅ **failed** case
> `charge.status = failed`, `charge.authorized = false`, `charge.paid = false`
alipay-failed

alipay-failed-checkout

### 3.2. Offline
Those payment methods that do not have redirection step after placing an order.
Payment method: `Bill Payment Tesco`, `PayNow`

Nothing change.
This pull request does not touch on those Offline payment methods.

### 3.3. Card
Credit and Debit card.

Nothing change.
Creating charge with Credit Card without 3-D Secure does not redirect buyers out, therefore the callback feature won't be triggered.

creditcard

### 3.4). Credit Card with 3-D Secure
Credit and Debit card with Omise 3-D Secure enabled account.

#### 3.4.1). Auth and Capture (charge.capture = `true`)

3.4.1.1). ✅ **Successful** case
> `charge.status = successful`, `charge.authorized = true`, `charge.paid = true`
creditcard-3ds-paid-successful

Buyer will be redirected back to the store "thank-you" page after finished the 3-D Secure authorization process (OTP).

3.4.1.2). ✅ **Failed** case
> `charge.status = failed`, `charge.authorized = false`, `charge.paid = false`
creditcard-3ds-paid-failed-checkout

creditcard-3ds-paid-failed

Buyer will be redirected back to the store "checkout" page if the OTP process is failed.

> **IMPORTANT NOTE**
> For Credit Card, an order status will stay as `pending-payment` even though the payment is failed (as originally it was designed to let buyer re-enter or change their card again – so the plugin does not mark an order as failed).

3.4.1.3). ✅ **Pending** case
> `charge.status = pending`, `charge.authorized = false`, `charge.paid = false`
creditcard-3ds-paid-pending-user

creditcard-3ds-paid-pending

> **IMPORTANT NOTE**
> We have never been received any of report showing that this case ever occurs. However, this is to prevent in case it happens.

#### 3.4.2). Auth Only (charge.capture = `false`)

3.4.2.1). ✅ **Successful** case
> `charge.status = pending`, `charge.authorized = true`, `charge.paid = false`
creditcard-3ds-auth-successful

For the failure and pending case, it will result the same as `Auth and Capture` cases.

## 4. Impact of the change

Some response messages have been changed, that would impact to the plugin translation.

## 5. Priority of change

Normal

## 6. Additional Notes

Nothing for now


includes/gateway/class-omise-payment-alipay.php
```diff
@@ -18,7 +18,7 @@ public function __construct() {
 			$this->description          = $this->get_option( 'description' );
 			$this->restricted_countries = array( 'TH' );
 
-			add_action( 'woocommerce_api_' . $this->id . '_callback', array( $this, 'callback' ) );
+			add_action( 'woocommerce_api_' . $this->id . '_callback', array( new Omise_Callback, 'execute' ) );
```
jonrandy, line 21: Seems to be no point to instantiating an object here. The 'callback' object has no state and is merely being used as a holder for some functions. Would probably be better to not instantiate an object, and just use static class methods

guzzilar: IIRC, I was trying static class but `add_action()` does not accept for a static class here.

jonrandy:
I looked that up. It does work apparently

includes/gateway/class-omise-payment-alipay.php
guzzilar: @jonrandy  Still researching, but wonder if there is any downside here by initiating a new object?

guzzilar: @jonrandy another approach I'm thinking to go is to turn this `Omise_Callback` class into singleton class. I think it will be better than change it to bunch of static methods as it would give the code more flexibility and extensible ability in case if there is any more complex logic that 3rd-party want to implement by themselves. (however, not in this PR. But that extensible can come in the future PR if needed)

includes/class-omise-callback.php
```diff
@@ -0,0 +1,178 @@
+<?php
```
jonrandy, line 1: I'm unconvinced that a singleton is the way to go here. I don't see any point to it

guzzilar: @jonrandy I've mentioned here https://github.com/omise/omise-woocommerce/pull/167#discussion_r427031831

jonrandy: Yes, I read that before. Still unconvinced

guzzilar: Can you explain a bit on what part that doesn't make you buy the idea?

guzzilar: (There are 3 approaches now, initiating object, singleton, or static.
I agreed that we don't need to initiate a new object every time and I have provided the reason why I go with singleton instead of state-less static class as there are some benefits that it can be achieved better than singleton. Maybe I missed some point that you see.)

jonrandy: The class is just being used as a holder for some utility functions - a helper class. It has no state, and nothing that needs initialising before. Instantiating it in any way - be that as a singleton, or a regular class instance - just seems completely superfluous

guzzilar: Ah yup, then that's the point I've considered. I think it would be a useful trade-off between state-less static class and extensibility for 3rd-party developers (as this project is an open source by its nature). 
It's understandable if you are not convinced by this approach but if it does not do any harm to the project, then let me see it through a bit. If at the end it does not be any useful, then let's change it to static. I'm still convinced with this approach as in WordPress and WooCommerce community nature.

jonrandy: Such static class helpers can be subclassed and extended easily

guzzilar: Would it be considered as a good code practice to extend a static class?

includes/class-omise-callback.php
```diff
@@ -0,0 +1,178 @@
...
+
+		$order   = $callback->retrieve_order();
+		$payment = $callback->retrieve_payment( $order );
+
+		$order->add_order_note( __( 'OMISE: Validating the payment result...', 'omise' ) );
+
+		try {
+			$charge = OmiseCharge::retrieve( $order->get_transaction_id() );
+
+			$resolving_method = strtolower( 'payment_' . $charge['status'] );
```
robinclart, line 38: Wouldn't it be better to use a switch case (thus creating an explicit whitelist) here instead of allowing "untrusted" strings to call methods

includes/class-omise-callback.php
```diff
@@ -0,0 +1,178 @@
+<?php
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * @since 4.0
+ */
+class Omise_Callback {
```
robinclart, line 8: Three methods in here take both an order and a charge. That's refactoring 101 for "here's your state".

So, `execute` should get the order, make a `new Callback` set the order and finally call a member function (`execute` like the static is fine I guess). This function finds the appropriate charge, then `switch` on the `$charge['status']` and decide which private function to call (because there's no point in making those public).

Benefit is that now, if you need to run this manually, you can skip the static execute (which uses `$_GET['order_id']` to find the order), and instead instantiate a callback, set the order and call `->execute()`.

includes/class-omise-callback.php
```diff
@@ -0,0 +1,178 @@
...
+		}
+
+		return self::$the_instance;
+	}
+
+	public static function execute() {
+		$callback = self::instance();
+
+		$order   = $callback->retrieve_order();
+		$payment = $callback->retrieve_payment( $order );
```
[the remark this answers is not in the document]
guzzilar: Ugh, thanks for catching. 
It was there from the original implementation approach.

includes/class-omise-callback.php
guzzilar: Hmm, it seems to me the same.
Using switch cases we will have to define an exact condition for its case anyway, which, relying on the same "untrusted" string to meet its condition.

With this, those that extend this class (in the future) can simply add `payment_expired` method to handle those expired charges without a need to modify this class directly.

Unless, there is something relating to security concern that I have missed it 🤔

robinclart: It might not be too bad at the moment, but you never know how the code will evolve in the future. And you won't be thinking about this bit of dynamic behavior which could lead to a catastrophic scenario.

The advice we have in other repos is that if the string is untrusted (basically it's not coming from the repo itself and instead is a string coming from the outside) don't use it to call methods dynamically.

The most likely person to implement that `expired` scenario is you. So I wouldn't worry too much about extensibility :troll:

The other reason I don't like using this is that then you're using an exception to do flow control instead of using the else branch of the switch to define what is unknown behavior.

guzzilar: @robinclart actually the reason we need `try ... catch` there is because `OmiseCharge::retrieve()` will throw an exception in case if it cannot retrieve a charge object from Omise API, or else we wouldn't be able to catch an error message from Omise API, `error` object. Without wrapping everything under `exception` flow, then the code could have turn into some mess like this
```php
try {
	$charge = OmiseCharge::retrieve( $order->get_transaction_id() );
} catch ( Exception $e ) {
	$order->add_order_note(
		sprintf(
			wp_kses( __( 'OMISE: Unable to validate the result.<br/>%s', 'omise' ), array( 'br' => array() ) ),
			$e->getMessage()
		)
	);

	$callback->invalid_result();
}

switch ($charge['status']) {
	case 'successful':
	case 'failed':
	case 'pending':
		$resolving_method = strtolower( 'payment_' . $charge['status'] );
		$callback->$resolving_method( $order, $charge );
		break;
	
	default:
		$callback->invalid_result();
		break;
}
```
 Which, it feels a bit irritating to see this mess... but, the point that you mentioned about using exception as flow control is a bad idea is interesting. Would you mind to explain about that a bit? (I've seen some debate over internet, but just wonder how do you view about it)

robinclart: Right. Read the code too fast. Strike my last point then.

includes/class-omise-callback.php
guzzilar: @robinclart Been thinking about this, and I like the idea of manually set order so we can skip the static execute if needed. But probably at this moment, where there is still no case (and never has a potential) that would need that approach, how about we just keep the code as simple as it is and let the code gets evolved when it's needed?
What do you think?

robinclart: Did you actually try this out and judged which one looked/read better?

robinclart: To me the code doesn't look simple as it is. That $order, $charge tuple you're passing around screams "refactor me". And the whitelist still looks more complicated than it should. I just would not do the dynamic function calling. The added benefit is that you can pick better names for your methods as well if you wish.

guzzilar: @robinclart was thinking and designing a bit but not completely finish the code
However, just for context, originally, I decided to come for this approach as I want to leave the code programatically hook-able for any 3rd-party developers that they want to alter the "result" of `return_uri` callback behaviour. 
i.e. in WordPress, developers often use `add_action()` or `add_filter()` functions to hook and alter a specific point of code. 
The same as the way we are able to hook our callback function to WC as in the following code
```php
add_action( 'woocommerce_api_omise_alipay_callback', 'Omise_Callback::execute' );
```

That, is also the reason why I was deciding to pass `$order` and `$charge` as an argument instead of class properties, so, in the future, if we want to open these methods to be able to be hooked, then we can update the core code a bit, and developers can simply implement, on their theme directory:
```php
// function.php
add_action( 'omise_callback_payment_successful', 'my_payment_successful', 0, 2 );
function my_payment_successful( $order, $charge ) {
    // ... then they can alter from this point to serve their desire website flow
}
```

Or even as for extending Omise_Callback class. As, it'd be easier that developers don't need to look up what properties the class has, but do whatever in a scope of method's arguments.

--
But, give me some time to design the code for your suggestion a bit. I will code something out.

robinclart: I don't think what you're saying and what I'm saying is incompatible.

includes/class-omise-callback.php
```diff
@@ -0,0 +1,172 @@
...
+	public static function execute() {
+		$callback = self::instance();
+
+		$order = $callback->retrieve_order();
+		$order->add_order_note( __( 'OMISE: Validating the payment result...', 'omise' ) );
+
+		try {
+			$charge = OmiseCharge::retrieve( $order->get_transaction_id() );
+
+			switch ( strtolower( $charge['status'] ) ) {
```
danfowler, line 36: There are two more potential charge statuses: `expired` and `reversed` https://www.omise.co/charges-api

includes/class-omise-callback.php
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	public static function execute() {
+		$order_id = isset( $_GET['order_id'] ) ? sanitize_text_field( $_GET['order_id'] ) : null;
+
+		$callback = new self( wc_get_order( $order_id ) );
+		$callback->validate();
+	}
+
+	public function validate() {
```
guzzilar, line 34: Not confident if this is a perfect name. Naming ideas are welcomed :D

danfowler: Seems fine to me

guzzilar:
Note: During the re-implementation, I started to think that probably `Omise_Callback` is not a good place to keep state of Order and Charge objects. `Omise_Callback` gives me a feeling of it's just a controller that has only a job of handling a coming request and redirecting users to a proper result, where the actual state should be implemented somewhere else (maybe `Omise_Payment_Validator` or `Omise_Payment_Response`). But that, probably would be too much for this time so I left this solution out.

includes/class-omise-callback.php
```diff
@@ -0,0 +1,160 @@
...
+		$this->order->add_order_note( __( 'OMISE: Validating the payment result...', 'omise' ) );
+
+		try {
+			$this->charge = OmiseCharge::retrieve( $this->order->get_transaction_id() );
+
+			switch ( strtolower( $this->charge['status'] ) ) {
+				case 'successful':
```
danfowler, line 41: I see you 👍'd my previous comment, but dropping this here again:

https://github.com/omise/omise-woocommerce/pull/167/files/e444e4a05d89bfa092d3d06dd754fdad8a33a568#r439242588

includes/class-omise-callback.php
```diff
@@ -0,0 +1,160 @@
...
+				case 'failed':
+				case 'pending':
+					$resolving_method = strtolower( 'payment_' . $this->charge['status'] );
+					$this->$resolving_method();
+					break;
```
guzzilar, line 46: I still didn't change this part as this helps organising the code better to me, comparing to 
```php
switch ( strtolower( $this->charge['status'] ) ) {
    case 'successful':
        $this->payment_successful();
        break;

    case 'failed':
        $this->payment_failed();
        break;

    case 'pending':
        $this->payment_pending();
        break;

...
```
But if everyone strongly suggests the above ☝️ over the current, then I'll give up 😂

danfowler: It seems fine to me

robinclart: How about moving the switch into its own `dispatch` function if you think it looks messy.

```php
$this->dispatch($this->charge['status'])
```

Also allows you to `$callback->dispatch('successful')` from the outside if you need that.

robinclart:
@guzzilar I'd tend to agree with you on separation of concern but I don't think it's worthwhile at this point.

includes/class-omise-callback.php
guzzilar: @danfowler thanks, I left those 2 statuses out as there is no particular case to handle at the moment. But noted for a future improve 👍

includes/class-omise-callback.php
guzzilar: Thanks for suggesting 👍 
I have just write some part of it, the dispatch idea sounds interesting. But probably I'll leave it as it is for now until the requirement of using `dispatch()` outside of Callback context comes.  

```php
public function validate() {
    $this->order->add_order_note( __( 'OMISE: Validating the payment result...', 'omise' ) );

    try {
        $this->charge = OmiseCharge::retrieve( $this->order->get_transaction_id() );
        $this->dispatch( $this->charge['status'] );
    } catch ( Exception $e ) {
        $this->invalid_result( $e->getMessage() );
    }
}

public function dispatch( $status ) {
    if ( ! in_array( $status, array( 'successful', 'failed', 'pending' ) ) ) {
        $this->invalid_result( __( 'Unrecognized Omise Charge status.', 'omise' ) );
    }

    $resolving_method = strtolower( 'payment_' . $this->charge['status'] );
    $this->$resolving_method();
}

/**
 * Resolving a case of undefined charge status
 */
protected function invalid_result( $order_note = null ) {
    if ( $order_note ) {
        $this->order->add_order_note(
            sprintf(
                wp_kses( __( 'OMISE: Unable to validate the result.<br/>%s', 'omise' ), array( 'br' => array() ) ),
                $order_note
            )
        );
    }

    $message = __(
        '<strong>We cannot validate your payment result:</strong><br/>
            Note that your payment may have already been processed.<br/>
            Please contact our support team if you have any questions.',
        'omise'
    );

    wc_add_notice( wp_kses( $message, array( 'br' => array(), 'strong' => array() ) ), 'error' );
    wp_redirect( wc_get_checkout_url() );
    exit;
}
```

PR #156, opened 2020-03-18
Localizing Omise API's error messages

guzzilar:
## 1. Objective

Omise WooCommerce supports for Japanese and Thai stores, however, some amount of messages haven't been localised properly and still stay as English (especially those from API response messages).

This pull request is to properly addressing all those messages that can't be translated due to the implementation of the code.

**Related information**:
Related issue(s): T2778 (internal ticket)

## 2. Description of change

- Introducing Omise_Localization class to handle Omise API localization (https://github.com/omise/omise-woocommerce/pull/156/commits/1d426abb2d9ec27923160e0c8dfdc2942e4358e1)
- Translating Omise API error messages using Omise_Localization. (https://github.com/omise/omise-woocommerce/pull/156/commits/a832f88e8c4fbed22954e7e7e51f0fe7701261e2)

## 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: v4.3.3.
- **WordPress**: v5.4.2
- **PHP version**: 7.3.3

**✏️ Details:**

First of all, we can check if the translation covers all the possible part of the code by searching `'failure_message'`.
Screen Shot 2563-08-18 at 23 33 22

Try make a payment with an error card to see if Omise_Localization is working properly.
Screen Shot 2563-08-18 at 23 48 38

## 4. Impact of the change

nothing

## 5. Priority of change

Normal

## 6. Additional Notes

The messages that this PR is implementing with, is not the entire of all the messages that the plugin may receive from Omise API. However, this is to provide a structure and some sample messages first. The improvement could be in the future if there is a specific error message that this PR hasn't covered.

PR #168, opened 2020-05-11
Reduce duplicated code from offline payment methods to a dedicated class

guzzilar:
## 1. Objective

This pull request is aiming to reduce duplicated code that was discovered during the implementation of API Localization #156 

To reduce a place where we have to apply the L10n for those API response messages.

**Related information**:
Related issue(s): T21114 (internal ticket)

## 2. Description of change

•  Implementing a dedicated class **`Omise_Payment_Offline`** for those offline payment methods to reduce duplicated code.

## 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: v4.0.1
- **WordPress**: v5.4.1
- **PHP version**: 7.3.3.

**✏️ Details:**

As we're refactoring the `charge` and `result` methods, we should be making sure that those 2 offline payment methods can be created properly.

1. Making sure that users can place an order using Bill Payment Tesco payment method.
  1.1. Charge can be created properly.
  bill-successful
  1.2. All the params are assigned properly (price, description, metadata)
  bill-successful-dashboard

2. Making sure that users can place an order using **PayNow** payment method.
  2.1. Charge can be created properly.
  paynow-successful
  2.2. All the params are assigned properly (price, description, metadata)
  paynow-successful-dashboard

## 4. Impact of the change

No

## 5. Priority of change

Normal

## 6. Additional Notes

This `Omise_Payment_Offline` class may get modified again once we switch back to work on #149  Konbini payment (as Konbini requires more parameters to be passed when creating a new charge)

### DIFF
--- a/includes/gateway/abstract-omise-payment-offline.php
+++ b/includes/gateway/abstract-omise-payment-offline.php
@@ -1,0 +1,63 @@
+<?php
+
+defined( 'ABSPATH' ) || exit;
+
+require_once dirname( __FILE__ ) . '/class-omise-payment.php';
+
+/**
+ * @since 4.0
+ */
+abstract class Omise_Payment_Offline extends Omise_Payment {
+	/**
+	 * A string of Omise Source's type
+	 * (expecting for either 'paynow' or 'bill_payment_tesco_lotus').
+	 *
+	 * @var string
+	 */
+	protected $source_type;
+
+	/**
+	 * @inheritdoc
+	 */
+	public function charge( $order_id, $order ) {
+		$total    = $order->get_total();
+		$currency = $order->get_order_currency();
+		$metadata = array_merge(
+			apply_filters( 'omise_charge_params_metadata', array(), $order ),
+			array( 'order_id' => $order_id ) // override order_id as a reference for webhook handlers.
+		);
+
+		return OmiseCharge::create( array(
+			'amount'      => Omise_Money::to_subunit( $total, $currency ),
+			'currency'    => $currency,
+			'description' => apply_filters( 'omise_charge_params_description', 'WooCommerce Order id ' . $order_id, $order ),
+			'source'      => array( 'type' => $this->source_type ),
+			'metadata'    => $metadata
+		) );
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function result( $order_id, $order, $charge ) {
+		if ( self::STATUS_FAILED === $charge['status'] ) {
+			return $this->payment_failed( $charge['failure_message'] . ' (code: ' . $charge['failure_code'] . ')' );
+		}
+
+		if ( self::STATUS_PENDING === $charge['status'] ) {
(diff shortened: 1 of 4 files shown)


includes/gateway/abstract-omise-payment-offline.php
```diff
@@ -0,0 +1,63 @@
...
+			array( 'order_id' => $order_id ) // override order_id as a reference for webhook handlers.
+		);
+
+		return OmiseCharge::create( array(
+			'amount'      => Omise_Money::to_subunit( $total, $currency ),
+			'currency'    => $currency,
+			'description' => apply_filters( 'omise_charge_params_description', 'WooCommerce Order id ' . $order_id, $order ),
+			'source'      => array( 'type' => $this->source_type ),
+			'metadata'    => $metadata
+		) );
```
guzzilar, line 36: Note that `return_uri` has been removed. It is because Offline payment methods don't require for the `return_uri` parameter.

includes/gateway/abstract-omise-payment-offline.php
```diff
@@ -0,0 +1,63 @@
...
+
+require_once dirname( __FILE__ ) . '/class-omise-payment.php';
+
+/**
+ * @since 4.0
+ */
+abstract class Omise_Payment_Offline extends Omise_Payment {
+	/**
+	 * A string of Omise Source's type
+	 * (expecting for either 'paynow' or 'bill_payment_tesco_lotus').
```
jonrandy, line 13: Comment should probably go as no doubt more offline payment methods will be added

guzzilar: @jonrandy any suggestion? (can't think of other message than adding a word `at the moment`).
```php
/**
 * A string of Omise Source's type
 * 'bill_payment_tesco_lotus', 'paynow' (at the moment).
 */
```

jonrandy: Just remove that line of the comment, the rest is fine

bigomise:
@guzzilar not sure that this PR will fix issue what if the merchant would like to use live mode, but the setting is automatically changed to test mode?

guzzilar:
@bigomise Can I have more detail about that case? Never heard that issue before.

includes/gateway/abstract-omise-payment-offline.php
```diff
@@ -0,0 +1,63 @@
...
+	public function charge( $order_id, $order ) {
+		$total    = $order->get_total();
+		$currency = $order->get_order_currency();
+		$metadata = array_merge(
+			apply_filters( 'omise_charge_params_metadata', array(), $order ),
+			array( 'order_id' => $order_id ) // override order_id as a reference for webhook handlers.
+		);
+
+		return OmiseCharge::create( array(
+			'amount'      => Omise_Money::to_subunit( $total, $currency ),
```
danfowler, line 31: In the future, we may want to consider passing custom `expires_at` params

guzzilar: Noted 👍

PR #172, opened 2020-06-24
Move usage docs to website

danfowler:
#### 1. Objective

Usage docs are at https://www.omise.co/woocommerce-plugin

#### 2. Description of change

Just the `README`

#### 3. Quality assurance

n/a

#### 4. Impact of the change

n/a

#### 5. Priority of change

Normal

#### 6. Additional Notes

n/a


guzzilar:
Would it be better to leave this Contributing section as it is? It's a different content from our Doc and more focus on guiding user where and how to submit issue and PR. 
Screen Shot 2563-06-25 at 10 19 35

danfowler:
@guzzilar thanks for the feedback, but I think most of the existing content is implicit to making contributions on GitHub in general.  We can start re-adding to this section once we have more specific guidelines for contributors.

Also, the license should be in a `LICENSE` file.  I've made a separate PR to add it: https://github.com/omise/omise-woocommerce/pull/173

PR #175, opened 2020-07-03
Revising Omise Settings

guzzilar:
### 1. Objective

There are still some more of UI that could enhance **Omise Setting** page.
This pull request is aiming to give an enhancement to it. On both UI and the code itself.

### 2. Description of change

2.1. Adding a **test-mode indicator** bar on top of the page.
Screen Shot 2563-07-02 at 17 19 40

2.2. Also, displaying an Omise account's email that is currently linked to a store.

2.3. Displaying notification messages when the setting form is submitted.

2.4. Fixing issue that Omise-WooCommerce test mode can't be disabled on WooCommerce version 4.

--
However, as for the beginning of the plugin (right after installed), everything will be the same.
Screen Shot 2563-07-02 at 17 38 10

### 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: v4.2.2.
- **WordPress**: v5.4.2.
- **PHP version**: 7.3.3.

**✏️ Details:**

In order to test all the cases, you may need a fresh-installed Omise-WooCommerce plugin.
3.1. Making sure that a fresh-installed Omise-WooCommerce plugin would function as the same.

Screen Shot 2563-07-04 at 15 59 40

3.2. Test input a wrong credential, Omise plugin should response as `authentication failed`
Screen Shot 2563-07-04 at 16 01 53

Screen Shot 2563-07-04 at 16 02 05

3.3. Test input test keys, the plugin should display a test-mode indicator at the top of the page.

Screen Shot 2563-07-04 at 16 09 07

3.4. Test enable live mode, the test-mode iindicator should be disappeared

Screen Shot 2563-07-04 at 16 11 16

#### 4. Impact of the change

none

#### 5. Priority of change

Normal

#### 6. Additional Notes

Nothing

### DIFF
--- a/includes/admin/class-omise-admin-page.php
+++ b/includes/admin/class-omise-admin-page.php
@@ -4,7 +4,7 @@ defined( 'ABSPATH' ) || exit;
 /**
  * @since 4.0
  */
-class Omise_Admin_Page {
+class Omise_Admin_Page extends Omise_Setting {
 	/**
 	 * @var array  a set of system messages.
 	 */
--- a/includes/admin/class-omise-admin-page.php
+++ b/includes/admin/class-omise-admin-page.php
@@ -7,12 +7,12 @@ defined( 'ABSPATH' ) || exit;
  */
 class Omise_Admin_Page extends Omise_Setting {
 	/**
-	 * @var array  a set of system messages.
+	 * @var array  of system messages.
 	 */
 	protected $messages = array();
 
 	/**
-	 * @var array  a set of error messages.
+	 * @var array  of error messages.
 	 */
 	protected $errors = array();
 
--- a/includes/admin/class-omise-admin-page.php
+++ b/includes/admin/class-omise-admin-page.php
@@ -1,4 +1,5 @@
 <?php
+
 defined( 'ABSPATH' ) || exit;
 
 /**
@@ -31,6 +32,9 @@ class Omise_Admin_Page extends Omise_Setting {
 		}
 	}
 
+	/**
+	 * @return HTML
+	 */
 	public function display_messages() {
 		if ( count( $this->errors ) > 0 ) {
 			foreach ( $this->errors as $error ) {
--- a/includes/admin/class-omise-admin-page.php
+++ b/includes/admin/class-omise-admin-page.php
@@ -1,0 +1,47 @@
+<?php
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * @since 4.0
+ */
+class Omise_Admin_Page {
+	/**
+	 * @var array  a set of system messages.
+	 */
+	protected $messages = array();
+
+	/**
+	 * @var array  a set of error messages.
+	 */
+	protected $errors = array();
+
+	/**
+	 * @param string $type     Whether 'error' or 'message'
+	 * @param string $message  A message to display
+	 */
+	public function add_message( $type, $message ) {
+		switch ( $type ) {
+			case 'error':
+				$this->errors[] = $message;
+				break;
+
+			case 'message':
+				$this->messages[] = $message;
+				break;
+		}
+	}
+
+	public function display_messages() {
+		if ( count( $this->errors ) > 0 ) {
+			foreach ( $this->errors as $error ) {
+				echo '<div class="error"><p>' . esc_html( $error ) . '</p></div>';
+			}
+		}
+
+		if ( count( $this->messages ) > 0 ) {
+			foreach ( $this->messages as $message ) {
+				echo '<div class="updated"><p>' . esc_html( $message ) . '</p></div>';
+			}
+		}
+	}
+}
--- a/includes/admin/views/omise-page-settings.php
+++ b/includes/admin/views/omise-page-settings.php
@@ -1,7 +1,21 @@
 <div class="wrap omise">
+	<style>
+		.omise-notice-testmode {
+			background: #ffce00;
+			color: #575D66;
+			border: 1px solid #efc200;
+			border-left-width: 4px;
+		}
+	</style>
 	<h1><?php echo $title; ?></h1>
 
 	<?php $page->display_messages(); ?>
+
+	<?php if ( 'yes' === $settings['sandbox'] && $settings['account_email'] ) : ?>
+		<div class="notice omise-notice-testmode">
+			<p><?php echo _e( 'You are in test mode. No actual payment is made in this mode', 'omise' ); ?></p>
+		</div>
+	<?php endif; ?>
 
 	<h2><?php echo _e( 'Payment Settings', 'omise' ); ?></h2>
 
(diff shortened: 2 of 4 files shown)


includes/admin/views/omise-page-settings.php
```diff
@@ -1,6 +1,24 @@
...
+			border-left-width: 4px;
+		}
+	</style>
 	<h1><?php echo $title; ?></h1>
+
+	<?php $page->display_messages(); ?>
+
+	<?php if ( 'yes' === $settings['sandbox'] && $settings['account_email'] ) : ?>
+		<div class="notice omise-notice-testmode">
+			<p><?php echo _e( 'You are in test mode. No actual payment is made in this mode', 'omise' ); ?></p>
```
jonrandy, line 16: Is there a corresponding similar warning at the checkout?

includes/admin/views/omise-page-settings.php
```diff
@@ -16,7 +34,26 @@
...
+						<td>
+							<fieldset>
+								Connected: <em><?php echo $settings['account_email']; ?> (<?php echo $settings['account_country']; ?>)</em>
+							</fieldset>
+						</td>
+					</tr>
+				</tbody>
+			</table>
+			<hr />
+		<?php endif; ?>
```
jonrandy, line 55: I've never understood the use of this syntax in 'templates'

includes/admin/views/omise-page-settings.php
guzzilar: For the test-mode warning bar, at the moment, no (maybe another PR)

includes/admin/views/omise-page-settings.php
guzzilar: I guess because WordPress has lots of HTML mixed with PHP,  so to make it easier to read. Or maybe it is to make those non-programmers to be able to read or to make their own theme more easy
```html
<?php if ( $a ) : ?>
    <?php while( $posts ) : $post = get_post(); ?>
        <div id="<?php echo $post->id; ?>">
            <div something>
                <?php if ( $a ) : ?>
                    <abel></label>
                <?php endif; ?>
            </div>
        </div>
    <?php endwhile; ?>
<?php endif; ?>
```

compare to

```html
<?php if ( $a ) { ?>
    <?php while( $posts ) { $post = get_post(); ?>        
        <div id="<?php echo $post->id; ?>">
            <div something>
                <?php if ( $b ) { ?>
                    <abel></label>
                <?php } ?>
            </div>
       </div>
    <?php } ?>
<?php } ?>
```

includes/admin/views/omise-page-settings.php
```diff
@@ -16,7 +34,26 @@
 		);
 		?>
 	</p>
+
 	<form method="POST">
+		<!-- Section: account information -->
+		<?php if ( $settings['account_email'] ) : ?>
```
danfowler, line 40: Why only check for `account_email` if you are also displaying `account_country`?

includes/admin/class-omise-admin-page.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * @since 4.0
+ */
+class Omise_Admin_Page extends Omise_Setting {
+	/**
+	 * @var array  a set of system messages.
```
danfowler, line 10: PHP also has a set ("a sequence of unique values") class, right?

includes/admin/views/omise-page-settings.php
```diff
@@ -1,6 +1,24 @@
...
+			color: #575D66;
+			border: 1px solid #efc200;
+			border-left-width: 4px;
+		}
+	</style>
 	<h1><?php echo $title; ?></h1>
+
+	<?php $page->display_messages(); ?>
+
+	<?php if ( 'yes' === $settings['sandbox'] && $settings['account_email'] ) : ?>
```
danfowler, line 14: Why check for `account_email` if the main thing we're concerned about is the value of `sandbox`?

guzzilar: Because the _sandbox_'s value is `yes` by default. Adding `$settings['account_email']` to make this bar display only when an Omise account has been set and the test mode is enabled.

includes/admin/views/omise-page-settings.php
bigomise: the first one seems to be more easier to read.

includes/admin/views/omise-page-settings.php
guzzilar: We may check `account_country` as well but I just think it is unnecessary.
This row will be displayed only when Omise account has been set, so I was looking for a variable that can be used as condition for this case. It will either have its value, or just empty string.
Default case:
```php
echo $settings['account_email']; // ""
echo $settings['account_country']; // ""
```

After connect with Omise account
```php
echo $settings['account_email']; // "...@omise.co"
echo $settings['account_country']; // "TH"
```

For the rare case (that I'm not sure how could it happen), it might be as this
```php
echo $settings['account_email']; // "...@omise.co"
echo $settings['account_country']; // ""
```
However, it's still safe to print `$settings['account_country']` out as the default of this variable is empty string.

includes/admin/views/omise-page-settings.php
```diff
@@ -16,7 +34,26 @@
...
 	</p>
+
 	<form method="POST">
+		<!-- Section: account information -->
+		<?php if ( $settings['account_email'] ) : ?>
+			<hr />
+			<table class="form-table">
+				<tbody>
+					<tr>
+						<th scope="row"><label><?php _e( 'Account status', 'omise' ); ?></label></th>
```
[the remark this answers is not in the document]
guzzilar: I was choosing `Account status` to support a case where that particular account get suspended then, instead of showing "connected", we can change its status to `blocked` or `suspended`. 

The output at this moment is:
> Account status: Connected: _user@omise.co (TH)_

The improvement will be
> Account status: Suspended: _user@omise.co (TH)_

includes/admin/views/omise-page-settings.php
danfowler: I suppose, but it's still true that no actual payment can be made when you haven't connected an account, right? 😏

includes/admin/views/omise-page-settings.php
danfowler: OK. I suppose in the theoretical case where we make some sort of breaking change which removes `country` from the `account` object, this isn't where we'd break anyway.

includes/admin/views/omise-page-settings.php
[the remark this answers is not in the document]
danfowler: I understand.

includes/admin/class-omise-admin-page.php
guzzilar: I missed this one. Will update to 
```php
/**
 * @var array  of system messages.
 ...
```

includes/admin/views/omise-page-settings.php
guzzilar: Hmm, good point actually

PR #177, opened 2020-07-14
Preventing async requests from Omise Webhook and Callback URI from updating WC Order status twice.

guzzilar:
## 1. Objective

In a race-condition where Omise Webhook is fired at the same time as buyer returns to the Callback URI (`return_uri`). A particular WooCommerce Order object will get updated its status twice which creates a consequence of a confirmation email is being sent to the recipient twice causing a confusion in a business flow that rely on this particular email.

Screen Shot 2563-07-15 at 01 52 59 copy

This happens because at the moment where Omise Webhook and Callback URI are triggered asynchronously at the same time, they both retrieve WC Order object with the `pending` status and update it to `processing` using `WC_Order::payment_complete()`, which triggers an email function to be executed twice. 

This pull request is providing a solution to prevent the async request from creating a race-condition, by integrating with WC_Queue feature to handle those async request by executing it by order.

**Related information**:
Related issue(s): T22125 (internal ticket)

## 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: v4.2.2.
- **WordPress**: v5.4.2
- **PHP version**: 7.3.3.

**✏️ Details:**

> ⚠️ At the moment, I have no other approach to test the race-condition but to try create and place an order again and again until the Webhook event is finally arrived to the store first before user get redirected back to the Callback URI.

> ⚠️ Also, to test this pull request, you may use `ngrok` to create HTTPS public url for Omise Webhook.

> ⚠️ To check if the Webhook is finally arrived first, you may query database with the following command:
> `SELECT * from wp_actionscheduler_actions;`
> Or `SELECT action_id, hook, status, extended_args from wp_actionscheduler_actions ORDER BY action_id DESC LIMIT 5;`
> Screen Shot 2563-07-15 at 04 58 52 copy
> In a typical case, you will get the action `omise_async_payment_result` with `context: callback` come first following by `context: webhook`. This means the Callback URI is being executed first (buyer arrives first).
> And another way around, if Webhook arrived first or at the same time as buyer.

Once everything is in placed and you have set Webhook properly, you may try to place an order again until the Webhook is finally arrived to the store first before the Callback URI being executed.

Screen Shot 2563-07-15 at 04 56 38

If you check at the Admin Order Detail page, there will no longer be a double customer note (meaning that the request does not reach to the `WC_Order::payment_complete()` function).
Screen Shot 2563-07-15 at 05 38 40 copy

## 4. Impact of the change

There is a limitation of concurrent that `Action Scheduler` library can handle.
Please check the following document as reference: https://actionscheduler.org/perf

> **Action Scheduler will only process actions in a request until:**
> • 90% of available memory is used
> • processing another 3 actions would exceed 30 seconds of total request time, based on the average processing time for the current batch
> • in a single concurrent queue

Also

> By default, Action Scheduler will only process actions for a maximum of 30 seconds in each request. This time limit minimises the risk of a script timeout on unknown hosting environments, some of which enforce 30 second timeouts.

> By default, Action Scheduler will claim a batch of 25 actions. This small batch size is because the default time limit is only 30 seconds; however, if you know your actions are processing very quickly, e.g. taking microseconds not seconds, or that you have more than 30 second available to process each batch, increasing the batch size can slightly improve performance.

> By default, Action Scheduler will run only one concurrent batches of actions. This is to prevent consuming a lot of available connections or processes on your webserver.

However, it shall not effect to its capability to handle the queue. Just for some high-volume transaction per minute website, it may take some seconds or minutes until all the queue get executed. 

As stated in WP-Cron: https://developer.wordpress.org/plugins/cron/#what-is-wp-cron

## 5. Priority of change

Immediate.

## 6. Additional Notes

Just to clarify the relationship between 3 names.
- **WooCommerce** integrated **Action Scheduler** library in the project
- **Action Scheduler** is designed to work with **WP-Cron**
- **WP-Cron** is a default built-in Cron worker that WordPress provided.

However, there is one point that we should understand of WP-Cron behaviour, to understand how the WC_Queue work.
As stated in WP-Cron document: https://developer.wordpress.org/plugins/cron/#what-is-wp-cron
> **WP-Cron** works by checking, on every page load, a list of scheduled tasks to see what needs to be run. Any tasks due to run will be called during that page load.

> **WP-Cron** does not run constantly as the system cron does; it is only triggered on page load.

> With the system scheduler, if the time passes and the task did not run, it will not be re-attempted. With WP-Cron, all scheduled tasks are put into a queue and will run at the next opportunity (meaning the next page load). So while you can’t be 100% sure when your task will run, you can be 100% sure that it will run eventually.

The WP-Cron is only working when "any" of page get loaded. It's not time-based as a typical system scheduler.
Meaning that, even though we set some schedule to be executed after "5" mins, but if there is no one open "any" page of the WordPress website, then that schedule won't be triggered.

### DIFF
--- a/includes/class-omise-callback.php
+++ b/includes/class-omise-callback.php
@@ -38,7 +38,7 @@ class Omise_Callback {
 	}
 
 	public function validate() {
-		$this->order->add_order_note( __( 'Omise: user is redirected back from the payment page.', 'omise' ) );
+		$this->order->add_order_note( __( 'Omise: User has been redirected back from the payment page.', 'omise' ) );
 
 		try {
 			$this->charge = OmiseCharge::retrieve( $this->order->get_transaction_id() );
--- a/includes/class-omise-callback.php
+++ b/includes/class-omise-callback.php
@@ -7,6 +7,11 @@ defined( 'ABSPATH' ) || exit;
  */
 class Omise_Callback {
 	/**
+	 * @var string
+	 */
+	protected $order_id;
+
+	/**
 	 * @var \WC_Abstract_Order
 	 */
 	protected $order;
@@ -19,20 +24,21 @@ class Omise_Callback {
 	/**
 	 * @param \WC_Abstract_Order $order
 	 */
-	public function __construct( $order ) {
-		$this->order = $order;
+	public function __construct( $order_id ) {
+		$this->order_id = $order_id;
+		$this->order    = wc_get_order( $order_id );
 		if ( ! $this->order || ! $this->order instanceof WC_Abstract_Order ) $this->invalid_result();
 	}
 
 	public static function execute() {
 		$order_id = isset( $_GET['order_id'] ) ? sanitize_text_field( $_GET['order_id'] ) : null;
 
-		$callback = new self( wc_get_order( $order_id ) );
+		$callback = new self( $order_id );
 		$callback->validate();
 	}
 
 	public function validate() {
-		$this->order->add_order_note( __( 'OMISE: Validating the payment result...', 'omise' ) );
+		$this->order->add_order_note( __( 'Omise: user is redirected back from the payment page.', 'omise' ) );
 
 		try {
 			$this->charge = OmiseCharge::retrieve( $this->order->get_transaction_id() );
@@ -41,6 +47,12 @@ class Omise_Callback {
 				case 'successful':
 				case 'failed':
 				case 'pending':
+					WC()->queue()->add(
+						'omise_async_payment_result',
+						array( 'order_id' => $this->order_id, 'charge' => serialize( $this->charge ), 'context' => 'callback' ),
+						'omise-payment-result'
+					);
+
 					$resolving_method = strtolower( 'payment_' . $this->charge['status'] );
 					$this->$resolving_method();
 					break;
@@ -81,17 +93,6 @@ class Omise_Callback {
 	 * Resolving a case of charge status: successful.
 	 */
 	protected function payment_successful() {
-		$message = __( 'OMISE: Payment successful.<br/>An amount of %1$s %2$s has been paid', 'omise' );
-
-		$this->order->payment_complete();
-		$this->order->add_order_note(
-			sprintf(
-				wp_kses( $message, array( 'br' => array() ) ),
-				$this->order->get_total(),
-				$this->order->get_order_currency()
-			)
-		);
-
 		WC()->cart->empty_cart();
 		wp_redirect( $this->order->get_checkout_order_received_url() );
 		exit;
@@ -101,38 +102,12 @@ class Omise_Callback {
 	 * Resolving a case of charge status: pending.
 	 */
 	protected function payment_pending() {
+		// Card authorized case.
 		if ( ! $this->charge['capture'] && $this->charge['authorized'] ) {
-			// Card authorized case.
-			$message = __(
-				'Omise: The payment is being processed.<br/>
-				 An amount %1$s %2$s has been authorized.',
-				'omise'
-			);
-
-			$this->order->add_order_note(
-				sprintf(
-					wp_kses( $message, array( 'br' => array() ) ),
-					$this->order->get_total(),
-					$this->order->get_order_currency()
-				)
-			);
-
 			// Remove cart
 			WC()->cart->empty_cart();
-			wp_redirect( $this->order->get_checkout_order_received_url() );
-			exit;
 		}
 
-		// Offsite case.
-		$message = __(
-			'Omise: The payment is being processed.<br/>
-			 Depending on the payment provider, this may take some time to process.<br/>
-			 Please do a manual \'Sync Payment Status\' action from the <strong>Order Actions</strong> panel, or check the payment status directly at the Omise Dashboard later.',
-			'omise'
-		);
-
-		$this->order->add_order_note( wp_kses( $message, array( 'br' => array(), 'strong' => array() ) ) );
-		$this->order->update_status( 'on-hold' );
 		wp_redirect( $this->order->get_checkout_order_received_url() );
 		exit;
 	}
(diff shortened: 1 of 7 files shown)


includes/class-omise-rest-webhooks-controller.php
```diff
@@ -42,14 +42,14 @@ public function callback( $request ) {
 			return new WP_Error( 'omise_rest_wrong_header', __( 'Wrong header type.', 'omise' ), array( 'status' => 400 ) );
 		}
 
-		$body = json_decode( $request->get_body() );
+		$body = json_decode( $request->get_body(), true );
```
guzzilar, line 45: Note: to make the payload data consistency on both Callback and Webhook data, this `json_decode` should be converting the payload data into array instead of object (as the same as OmiseCharge object).

This would help `Omise_Payment_Result` class to handle the same-format data.

includes/events/class-omise-event-charge-complete.php
```diff
@@ -44,68 +44,26 @@ class Omise_Event_Charge_Complete {
...
 		}
 
-		$order->add_order_note(
-			__(
-				'Omise: an event charge.complete has been caught (webhook).',
-				'omise'
-			)
-		);
-
-		switch ($data->status) {
```
guzzilar, line 67: Note: moving the entire order-status-manipulation logic to a dedicated class, `Omise_Payment_Result`. As the Callback and the Webhook event are sharing the same code.

includes/class-omise-callback.php
```diff
@@ -81,17 +93,6 @@ protected function invalid_result() {
...
-
-		$this->order->payment_complete();
-		$this->order->add_order_note(
-			sprintf(
-				wp_kses( $message, array( 'br' => array() ) ),
-				$this->order->get_total(),
-				$this->order->get_order_currency()
-			)
-		);
-
```
guzzilar, line 94: Note: moving all the code that is used for manipulating the order status into `Omise_Payment_Result` class. Leaving only those UI-related code here in the `Omise_Callback` class. 

As in a concept of, `Omise_Callback` would be only a route for buyer to be redirected back, then get resolved to a proper page based on the result of a particular charge.

includes/class-omise-payment-result.php
```diff
@@ -0,0 +1,122 @@
+<?php
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * @since 4.0
+ */
+class Omise_Payment_Result {
```
guzzilar, line 8: Note: still 50/50 of this classname. Should it be `Omise_Order_Result` instead? Since this class is mainly designed to be used for resolving a payment and properly updating a particular order status accordingly.

guzzilar:
I'll write out some more test cases this evening for those failure, pending, authorized cases.

includes/events/class-omise-event-charge-complete.php
```diff
@@ -44,68 +44,26 @@ class Omise_Event_Charge_Complete {
 	 * @return void
 	 */
 	public function handle( $data ) {
-		if ( 'charge' !== $data->object || ! isset( $data->metadata->order_id ) ) {
+		if ( 'charge' !== $data['object'] || ! isset( $data['metadata']['order_id'] ) ) {
```
mayurkathale, line 47: Same has to be done for other event classes i.e. `Omise_Event_Charge_Create` , ` Omise_Event_Charge_Capture`.

guzzilar: @mayurkathale awesome catch! 👍 
Done

includes/class-omise-payment-result.php
```diff
@@ -0,0 +1,129 @@
...
+
+		$this->order->add_order_note( $message );
+		$this->order->update_status( 'on-hold' );
+	}
+
+	/**
+	 * Resolving a case of an authorized charge.
+	 */
+	public function payment_authorized() {
+		$message = wp_kses( __(
```
mayurkathale, line 120: If manual capture is selected, this may cause adding duplicate note to order for charge authorisation. Similar note in `Omise_Payment_Creditcard`.

includes/class-omise-callback.php
```diff
@@ -81,17 +93,6 @@ protected function invalid_result() {
...
 		WC()->cart->empty_cart();
 		wp_redirect( $this->order->get_checkout_order_received_url() );
```
mayurkathale, line 97: Order status is still failed while redirecting to return page.

  When customer creates order with a failed charge and then create new charge using another payment method, then order get placed normally but it returns to order failure page showing payment error message.

guzzilar: @mayurkathale Thanks for testing this case. Can I know the steps you have tested? I wanna try reproduce it.

includes/class-omise-payment-result.php
guzzilar: @mayurkathale Can I have more detail about this case? 
Here is the manual-capture case.
Screen Shot 2563-07-15 at 16 16 29

includes/class-omise-callback.php
mayurkathale: 1. Place order using "Alipay".
2. On test payment page, mark payment as failed.
3. Step 2 will take you to checkout page again with failure message error in the notification box.
4. Try to placing same order again with Card Payment with successful charge creation.
5. Step 4 will redirect you to order complete page.
6. Order completed page shows that order placement has been failed.

While backend and omise backend shows payment has done successfully, email gets sent normally.

includes/class-omise-payment-result.php
mayurkathale: @guzzilar maybe need to do some activity on the website which might help to worker jobs to get executed. This should be done before payment capture.

includes/class-omise-callback.php
guzzilar: Oh, this is interesting case. Let me take a look

guzzilar: I couldn't reproduce this case (on both 3-D Secure payment and without).
Can you help test one more time? Maybe it's about browser session cache or something?

includes/class-omise-payment-result.php
guzzilar: I understand the issue now.
Just pushed a new commit here: https://github.com/omise/omise-woocommerce/pull/177/commits/84d09ab068eb94e0103514d5293b608d60390f6f.
Thanks for catching it 👍👍

guzzilar:
Not sure if this PR is concerning too many areas. Let me know if it'd be better to split some changes out to another different PR. (i.e. the Webhook's payload format from Object to array)

danfowler:
I think fine to keep as one.

includes/class-omise-callback.php
mayurkathale: @guzzilar I have tried in 'incognito mode', I can reproduce this issue. Probably, after step _number 3_, have to make sure that order status is "Failed" and try to execute worker job by doing some activity.

guzzilar: ok, I can reproduce the issue now.
Thanks 👍

guzzilar:
A new approach has been provided at #179

PR #185, opened 2020-08-17
PayNow, adding a link to the QR code url at the order-confirmation email

guzzilar:
## 1. Objective

Currently when placing an order using PayNow payment method, the QR code will be shown only on the order-received (thank-you) page.

This pull request is adding the QR code's link to the WC order-confirmation email as well.

**Related information**:
Related issue(s): T22531 (internal ticket)

## 2. Description of change

Adding a QR code link to the order-confirmation email when placing a new order using PayNow payment method.

## 3. Quality assurance

**🔧 Environments:**

- **WooCommerce**: v4.3.0
- **WordPress**: v5.4.2
- **PHP version**: 7.3.3

**✏️ Details:**

An order-confirmation email that is sent by placing a new order using PayNow payment method should contain a link to the PayNow QR code.
Screen Shot 2563-08-18 at 05 03 43

## 4. Impact of the change

Nothing

## 5. Priority of change

Normal

## 6. Additional Notes

None

### DIFF
--- a/includes/gateway/class-omise-payment-paynow.php
+++ b/includes/gateway/class-omise-payment-paynow.php
@@ -27,6 +27,7 @@ class Omise_Payment_Paynow extends Omise_Payment_Offline {
 		add_action( 'woocommerce_update_options_payment_gateways_' . $this->id, array( $this, 'process_admin_options' ) );
 		add_action( 'woocommerce_order_action_' . $this->id . '_sync_payment', array( $this, 'sync_payment' ) );
 		add_action( 'woocommerce_thankyou_' . $this->id, array( $this, 'display_qrcode' ) );
+		add_action( 'woocommerce_email_after_order_table', array( $this, 'email_qrcode' ) );
 	}
 
 	/**
@@ -59,6 +60,18 @@ class Omise_Payment_Paynow extends Omise_Payment_Offline {
 	}
 
 	/**
+	 * @param WC_Order $order
+	 *
+	 * @see   woocommerce/templates/emails/email-order-details.php
+	 * @see   woocommerce/templates/emails/plain/email-order-details.php
+	 */
+	public function email_qrcode( $order ) {
+		if ( $this->id == $order->get_payment_method() ) {
+			$this->display_qrcode( $order, 'email' );
+		}
+	}
+
+	/**
 	 * @param int|WC_Order $order
 	 * @param string       $context  pass 'email' value through this argument only for 'sending out an email' case.
 	 */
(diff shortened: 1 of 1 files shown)


includes/gateway/class-omise-payment-paynow.php
```diff
@@ -70,16 +83,19 @@ public function display_qrcode( $order, $context = 'view' ) {
...
-		?>
-		<div class="omise omise-paynow-details" <?php echo 'email' === $context ? 'style="margin-bottom: 4em; text-align:center;"' : ''; ?>>
-			<div class="omise omise-paynow-logo"></div>
-			<p>
-				<?php echo __( 'Scan the QR code to pay', 'omise' ); ?>
-			</p>
-			<div class="omise omise-paynow-qrcode">
-				<img src="<?php echo $qrcode; ?>" alt="Omise QR code ID: <?php echo $charge['source']['scannable_code']['image']['id']; ?>">
+
+		if ( 'view' === $context ) : ?>
```
mayurkathale, line 87: We should exclude this part if order placement has failed.

includes/gateway/class-omise-payment-paynow.php
```diff
@@ -70,16 +83,19 @@ public function display_qrcode( $order, $context = 'view' ) {
...
+			<div class="omise omise-paynow-details" <?php echo 'email' === $context ? 'style="margin-bottom: 4em; text-align:center;"' : ''; ?>>
+				<div class="omise omise-paynow-logo"></div>
+				<p>
+					<?php echo __( 'Scan the QR code to pay', 'omise' ); ?>
+				</p>
+				<div class="omise omise-paynow-qrcode">
+					<img src="<?php echo $qrcode; ?>" alt="Omise QR code ID: <?php echo $charge['source']['scannable_code']['image']['id']; ?>">
```
mayurkathale, line 94: This link downloads the QR image directly which is likely isn't standard flow. Should we create a public link to new page and display QR?

mayurkathale (changes requested):
Few comments above.

includes/gateway/class-omise-payment-paynow.php
guzzilar: I actually was 50/50 about this... One side kinda assume that to pay a QR code, probably buyer does not need to open an entire website to scan a code (also it's easier to implement). Another side, for its credibility, maybe we need to render the webpage.

What do you think?

includes/gateway/class-omise-payment-paynow.php
guzzilar: Am I understanding correctly that you mean when buyer attempts to scan the QR code and get `failed` status back from Charge API?

mayurkathale: sorry for the confusion, I wanted to add this comment into other part. consider this for line no 97.

includes/gateway/class-omise-payment-paynow.php
```diff
@@ -70,16 +83,19 @@ public function display_qrcode( $order, $context = 'view' ) {
...
+				</div>
 			</div>
-		</div>
-		<?php
+		<?php elseif ( 'email' === $context ) : ?>
```
mayurkathale, line 97: This should be excluded from email content if the charge with failed status.

guzzilar: From the implementation, buyers will not get into this part of code (thank-you page) if the charge status is failed.
Did you get the result otherwise? (maybe I missed some cases...)

danfowler (approved):

mayurkathale (approved):

PR #193, opened 2020-10-08
Add installment SCB support

nlogx:
#### 1. Objective

Add installment SCB support

#### 2. Description of change

- Added installment SCB to backends

#### 3. Quality assurance

**Works with Omise core#6704**

Test both test and live charges, make sure Installment SCB can be selected on checkout page:

Screen Shot 2020-10-08 at 17 41 09

Test result:

Screen Shot 2020-10-09 at 14 55 20

#### 4. Impact of the change

Supports installment SCB.

#### 5. Priority of change

High.

### DIFF
--- a/includes/backends/class-omise-backend-installment.php
+++ b/includes/backends/class-omise-backend-installment.php
@@ -53,6 +53,13 @@ class Omise_Backend_Installment extends Omise_Backend {
 				'interest_rate'      => 0.65,
 				'min_allowed_amount' => 500.00,
 			),
+
+			'installment_scb' => array(
+				'bank_code'          => 'scb',
+				'title'              => __( 'Siam Commercial Bank', 'omise' ),
+				'interest_rate'      => 0.74,
+				'min_allowed_amount' => 500.00,
+			),
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


alan-hadyk:
It works, but unfortunately payment method title is not visible for me:
Screen Shot 2020-10-08 at 1 04 31 PM

includes/backends/class-omise-backend-installment.php
```diff
@@ -53,6 +53,13 @@ public function initiate() {
 				'interest_rate'      => 0.65,
 				'min_allowed_amount' => 500.00,
 			),
+
+			'installment_scb' => array(
```
alan-hadyk, line 57: Maybe it would be a good idea to add a test case for this in `tests/unit/includes/backends/class-omise-backend-installment-test.php` ?

alan-hadyk (changes requested):

includes/backends/class-omise-backend-installment.php
nlogx: @alan-hadyk Added in https://github.com/omise/omise-woocommerce/pull/193/commits/9ece19785b0241f17a870d19fd3f65cb24f47c4d.

alan-hadyk (approved):

ratchagarn (approved):

mayurkathale (approved):

PR #203, opened 2020-12-16
Run tests using GitHub Actions

danfowler:
Run tests using GitHub Actions

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -1,0 +1,37 @@
+name: PHP Composer
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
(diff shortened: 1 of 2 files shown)


.github/workflows/php.yml
```diff
@@ -0,0 +1,37 @@
+name: PHP Composer
+
+on:
+  push:
```
mayurkathale, line 4: Can we remove "on push" keeping only "on pull request".

mayurkathale (changes requested):
notes above.

.github/workflows/php.yml
danfowler: Sure, updated.

mayurkathale (approved):

PR #222, opened 2021-07-13
[EM-611] Unable to view PromptPay QR after downloading from email

kan98:
#### 1. Objective

Explain in non-technical terms **WHY this PR is required**.
E.g.: What feature it adds, what problem it solves...

This section will be used in the release notes. 

When a customer creates an order and chooses Promptpay as their payment method, we send an order received in the email. The email includes the link to download the QR code to pay with, problem is the image is in an SVG format which isn't an ideal for customers to pay with. For many systems, there default application to open SVG images are a text editor to show it's XML.

**Related information**:
Related issue(s): #< GitHub ticket number > (optional)

#### 2. Description of change

A general description of **WHAT changed in the codebase**, but short of an English version of the diff. Assume that people reading this will also be looking at the output of `git diff` and guide them to the highlights.

Additionally add the reasoning for change details if they're complex or abstract.

Firstly, the email links to the order received woocommerce page instead of the image download link. Then in that page, we add a download image link to download the image as an SVG.

We add a script that runs on the page to create a canvas from the SVG and then create a PNG download link from that canvas. If the browser doesn't support the use of canvas, we will fallback to downloading the original SVG image instead. However, [most modern browsers supports canvas](https://caniuse.com/canvas) so users should be able to get the PNG image. 

#### 3. Quality assurance

Specify where and how you tested this and what further testing it might need.

Tested on Chrome, Firefox and Safari on Mac OS. Since we have a fallback to SVG and all the tests I conducted passed, I think it's OK to not test on more browsers.

**🔧 Environments:**

i.e.
- **WooCommerce**: v5.4.1
- **WordPress**: v5.7.2
- **PHP version**: 7.4

**✏️ Details:**

Explain how to manually test this feature.
For example if changes were made in the UI or in the API, explain where and if any specific access is needed.

Create a Promptpay charge on test/live mode and download the PNG. Also go to your email and click the link to come back to the order page and download the PNG.

#### 4. Impact of the change

List the steps that must be taken for this PR to work.
E.g.: rake yak:shave, Add "yak_key" to environment variables, ...

Be sure to include all systems that needs to be changed or which system is affected by the change
(Ex: Requires Elastic search to be installed and configured in secrets.yml).

Note: Please provide a screenshot if your changed impact to UI.

Screen Shot 2564-07-13 at 17 00 31

[qr_code.png](https://user-images.githubusercontent.com/34004612/125432913-7999bfe1-8083-47d0-9881-8ff544bd8ecd.png)

#### 5. Priority of change

Normal, High or Immediate.

Normal

#### 6. Additional Notes

Any further information that you would like to add.

### DIFF
--- a/assets/css/omise-css.css
+++ b/assets/css/omise-css.css
@@ -115,6 +115,31 @@ fieldset.card-exists {
 }
 
 /**
+* Omise Promptpay
+*/
+@media all and (min-width: 768px) {
+	.omise-promptpay-qrcode > img {
+		max-width: 300px !important;
+	}
+}
+
+.omise-download-promptpay-qr {
+	display: inline-block;
+	color: white !important;
+	padding: 6px 12px;
+	font-weight: bold;
+	background-color: #1A53F0;
+	border-radius: 4px;
+	margin-bottom: 0.75rem;
+	cursor: pointer;
+}
+
+.omise-download-promptpay-qr:hover {
+	background-color: #0c3abb;
+}
+
+
+/**
  * 2). Components
  * (stylesheets that are designed generally and may be commonly used as a component).
  */
--- a/assets/css/omise-css.css
+++ b/assets/css/omise-css.css
@@ -117,8 +117,14 @@ fieldset.card-exists {
 /**
 * Omise Promptpay
 */
+
+.omise-qr-image > svg {
+	max-width: 100%;
+	height: auto;
+}
+
 @media all and (min-width: 768px) {
-	.omise-promptpay-qrcode > img {
+	.omise-qr-image > svg {
 		max-width: 300px !important;
 	}
 }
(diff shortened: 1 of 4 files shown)


Ashraf-alkibsi (approved):

tookaa2 (approved):

som-m (approved):
Manually tested, all good. The link works with custom checkout endpoints as well.

One question though, how long is the order received page valid for?

kan98:
> Manually tested, all good. The link works with custom checkout endpoints as well.
> 
> One question though, how long is the order received page valid for?

The thank you page will never expire by default. It would be good to have as a reference for customers. If the payment is made from an account and not by a guest, it should only be available when the user is logged in. These are all WooCommerce's policies.

Our plugin wise, we only show the QR code and download button if the charge is still pending on our side. So if the charge is successful or failed, we won't show the QR code on the thank you page.

hpariwat (approved):

danfowler (approved):

PR #264, opened 2022-04-28
[FRON-1433]: Issue of PromptPay QR code download button not downloadi…

aashishgurung:
#### 1. Objective

Fix the issue of PromptPay QR code download button downloading an empty file.

**Jira Ticket**:
https://omise.atlassian.net/browse/FRON-1433

#### 2. Description of change

The query selector was pulling the first `svg` from the document. There were more than one `svg` element in the document and the intended `svg` was not the first one. This resulted in selecting a `svg` element which had no content. So, the download button downloaded a file with no content.

A second parameter `id` is added the `load_qr_svg_to_DOM` method to optionally add the id into the `svg` element. The default value will be null for `id`. This will help to query the specific element. The query selector is changed in the JS to target the specific svg element.

#### 3. Quality assurance

- Add an item to the card
- Go to checkout page
- Select PromptPay as the payment method
- Place order
- Click on Download QR

You should see a PNG file in your local machine. You should see a QR code opening the file.

**🔧 Environments:**

WooCommerce: v6.4.1
WordPress: v5.9.3
PHP version: 7.4.28
Omise plugin version: Omise-WooCommerce 4.19.2

### DIFF
--- a/assets/javascripts/omise-download-promptpay-as-png.js
+++ b/assets/javascripts/omise-download-promptpay-as-png.js
@@ -4,7 +4,7 @@
     $('a#omise-download-promptpay-qr').click(function(e) {
         if (isCanvasSupported()) {
             e.preventDefault();
-            var svg = document.querySelector('svg');
+            var svg = document.getElementById('omise-promptpay-qrcode-svg');
 
             /*
             Because of a Webkit (Safari) bug, where it won't fetch images from the SVG in time the first time around,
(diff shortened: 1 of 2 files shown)


kan98 (approved):
nicee! thanks for fixing 👍🏽

new4762 (approved):

hpariwat (approved):

tanawin-opn:
test and aprroved krub.

PR #266, opened 2022-05-07
Identify card type from card number

radiant7766:
I am trying to identify card number to display the credit card processing company icon on the side of the field. So far I can identify the card number now but I can't display the icon on the side. It only returns the company name which does not work. Can you help me out?


aashishgurung:
Please open a new issue instead of a PR.

PR #257, opened 2022-04-04
[APM2-625] update assets for mobile banking logo

tookaa2:
#### 1. Objective

Update logos to match mobile banking app

**Related information**:
Related issue(s): #< GitHub ticket number > (optional)

#### 2. Description of change

Updated logos and labels according to request from product department
Screen Shot 2565-04-04 at 11 56 41

#### 3. Quality assurance

Tested on staging api to ensure charge can be created

**🔧 Environments:**

Specify the details of your test environments, including, for each, the platform version (on which the plugin was run), the Omise plugin version, and the versions of your system software such as PHP or Ruby.

i.e.
- **WooCommerce**: v4.3.0
- **WordPress**: v5.4.2
- **PHP version**: 7.3.3
- **Omise plugin version**: Omise-WooCommerce 4.3 (optional, in case of submitting a new issue)

**✏️ Details:**

Explain how to manually test this feature.
For example if changes were made in the UI or in the API, explain where and if any specific access is needed.

#### 4. Impact of the change

List the steps that must be taken for this PR to work.
E.g.: rake yak:shave, Add "yak_key" to environment variables, ...

Be sure to include all systems that needs to be changed or which system is affected by the change
(Ex: Requires Elastic search to be installed and configured in secrets.yml).

Note: Please provide a screenshot if your changed impact to UI.

#### 5. Priority of change

Normal

#### 6. Additional Notes

Any further information that you would like to add.

PR #284, opened 2022-07-06
[PLUGINS-257: Sec-fetch header is checked to determine whether the re…

aashishgurung:
#### 1. Objective

Prevent customer to reach return_uri before success payment

Jira: [#257](https://opn-ooo.atlassian.net/browse/SHOPIFYAPP-257)

#### 2. Description of change

Sec-fetch header is checked to determine whether the request is a user originated operation or not. Created a new helper class request for it. This can be used to add request related helper function in the future.

**Reference**
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-Fetch-Site
- https://stackoverflow.com/questions/66115532/what-does-the-sec-fetch-site-header-mean-why-is-the-origin-header-undefined

HTTP_REFERER header was not present this time. We might need to change the implementation in Magento.

 **Reference**
- https://stackoverflow.com/questions/5551094/http-referer-coming-back-with-null-key-does-not-exist-in-server

#### 3. Quality assurance

- Setup to use manual capture
- Checkout with some items via Card.
- You will be redirected to the authorize page
  - Do nothing on authorize page. Open a new tab
- Go to the return URI `http://{YOUR_BASE_PATH}/?wc-api=omise_callback&order_id={ORDER_ID}`
- You should be redirected to the checkout page.
- Go to the authorize page and complete the transaction
- You should be redirected to the thank you page.

**🔧 Environments:**

- WooCommerce: v6.4.1
- WordPress: v5.9.3
- PHP version: 7.3.33
- Omise plugin version: Omise-WooCommerce 4.22.0

### DIFF
--- a/includes/class-omise-callback.php
+++ b/includes/class-omise-callback.php
@@ -25,6 +25,10 @@ class Omise_Callback {
 	}
 
 	public static function execute() {
+		if(RequestHelper::isUserOriginated()) {
+			return wp_redirect( wc_get_checkout_url() );
+		}
+
 		$order_id = isset( $_GET['order_id'] ) ? sanitize_text_field( $_GET['order_id'] ) : null;
 
 		$callback = new self( wc_get_order( $order_id ) );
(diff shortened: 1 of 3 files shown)


hpariwat (approved):

tanawin-opn:
@aashishgurung ✅ works as expected krub. it's also blocked the second cart page (when it redirect from return_uri) to check out after the first checkout page is already done.

kan98 (approved):

new4762 (approved):

PR #305, opened 2022-09-05
[ENGA3-420]: Updated Omise_Capabilities class to prevent calling capabilities API when the keys are null.

aashishgurung:
#### 1. Objective

Prevent calling capabilities  API when the keys are null.

Jira Ticket: [#420](https://opn-ooo.atlassian.net/browse/ENGA3-420)

#### 2. Description of change

Updated `Omise_Capabilities` class to prevent calling capabilities API when the keys are null. Added a check that compares the newly fetched keys with the previously saved keys. If the keys are different then then we call to capabilities API else we return previously built instance.

**🔧 Environments:**

- WooCommerce: v6.8.0
- WordPress: v6.0.2
- PHP version: 8.1
- Omise plugin version: Omise-WooCommerce 4.23.3

### DIFF
--- a/includes/class-omise-capabilities.php
+++ b/includes/class-omise-capabilities.php
@@ -56,7 +56,8 @@ class Omise_Capabilities {
 		try {
 			$capabilities = OmiseCapabilities::retrieve( $publicKey , $secretKey );
 		} catch(\Exception $e) {
-			// suppressing error on the admin dashboard
+			// logging the error and suppressing error on the admin dashboard
+			error_log(print_r($e, true));
 			return null;
 		}
 
--- a/includes/class-omise-capabilities.php
+++ b/includes/class-omise-capabilities.php
@@ -40,14 +40,16 @@ class Omise_Capabilities {
 		$secretKey = !$sKey ? $settings->secret_key() : $sKey;
 
 		// Do not call capabilities API if keys are not present
-		if($publicKey === null || $secretKey === null) {
+		if(empty($publicKey) || empty($secretKey)) {
 			return null;
 		}
 
 		if(self::$instance) {
 			$keysNotChanged = self::$instance->publicKey === $publicKey && self::$instance->secretKey === $secretKey;
 
-			// if keys are same then no need to call capabilities API as t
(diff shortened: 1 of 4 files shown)


new4762 (approved):
Thank you for your hard work. 🙇

includes/class-omise-capabilities.php
```diff
@@ -8,30 +8,63 @@ class Omise_Capabilities {
...
+
+		// Do not call capabilities API if keys are not present
+		if($publicKey === null || $secretKey === null) {
+			return null;
+		}
+
+		if(self::$instance) {
+			$keysNotChanged = self::$instance->publicKey === $publicKey && self::$instance->secretKey === $secretKey;
+
+			// if keys are same then no need to call capabilities API as t
```
ajzkk, line 50: did you forget to complete comment?

includes/class-omise-capabilities.php
```diff
@@ -8,30 +8,63 @@ class Omise_Capabilities {
...
-				// suppressing error on the admin dashboard
-				return null;
+	public static function retrieve($pKey = null, $sKey = null)
+	{
+		$settings = Omise_Setting::instance();
+		$publicKey = !$pKey ? $settings->public_key() : $pKey;
+		$secretKey = !$sKey ? $settings->secret_key() : $sKey;
+
+		// Do not call capabilities API if keys are not present
+		if($publicKey === null || $secretKey === null) {
```
ajzkk, line 43: Better to use `empty()` instead checking null?

ajzkk (approved):

includes/class-omise-capabilities.php
aashishgurung: @ajzkk No idea what happened here. I have add more context. Thanks.

includes/class-omise-capabilities.php
aashishgurung: @ajzkk Right. It better to check for any falsy value rather than just null. Thanks.

kan98 (approved):

FhanOmise (approved):

PR #306, opened 2022-09-06
[Feature/ENGA3-423]: Fix the issue of mobile banking not return to success/failed page.

aashishgurung:
#### 1. Objective

Fix the issue of mobile banking not return to success/failed page.

Jira Ticket: [#423](https://opn-ooo.atlassian.net/browse/ENGA3-423)

#### 2. Description of change

Change the callback URI from an endpoint to the query parameters.

#### 3. Quality assurance

Currently, we cannot test in staging as all testing apps are disabled.

**🔧 Environments:**

- WooCommerce: v6.8.0
- WordPress: v6.0.2
- PHP version: 8.1
- Omise plugin version: Omise-WooCommerce 4.23.3


kan98 (approved):

includes/gateway/class-omise-payment-mobilebanking.php
```diff
@@ -81,7 +81,10 @@ public function charge( $order_id, $order ) {
 		);
 
 		$source_type = sanitize_text_field( $_POST['omise-offsite']);
-		$return_uri = add_query_arg('order_id', $order_id, home_url('wc-api/omise_mobilebanking_callback'));
```
new4762, line 84: so the root cause it bc we can't consturct by using this format right? 
 then why we roll back to 4.22.2 and it work na 🤔

new4762 (approved):

FhanOmise (approved):

aashishgurung:
Closing this PR because the issue is not due to our code and caused due to the settings in the permalinks of the merchant which has already been fixed.

PR #318, opened 2022-10-03
[ENGA3-448]: Customer can saved the card if the customer ID is remove…

aashishgurung:
#### 1. Objective

Customer can saved the card if the customer ID is removed separately from the API.

Jira Ticket: [#448](https://opn-ooo.atlassian.net/browse/ENGA3-448)

#### 2. Description of change

Before charging, we check if the saved customer ID exists in Omise server or not. If not then we create a new customer and save the customer ID to the Wordpress.

#### 3. Quality assurance

- Create new account on Wordpress/WooCommerce
- Checkout with 3DS and save card when making a payment
- Delete that customer ID from Omise via API
- Try to use the same Wordpress account to save card again and it should not fail

**🔧 Environments:**
- WooCommerce: v6.8.2
- WordPress: v6.0.2
- PHP version: 8.1
- OmisePHP version: 2.16.0

### DIFF
--- a/includes/libraries/omise-php/lib/omise/res/OmiseApiResource.php
+++ b/includes/libraries/omise-php/lib/omise/res/OmiseApiResource.php
@@ -1,8 +1,8 @@
 <?php
 
 define('OMISE_PHP_LIB_VERSION', '2.16.0');
-define('OMISE_API_URL', 'https://api.staging-omise.co/');
-define('OMISE_VAULT_URL', 'https://vault.staging-omise.co/');
+define('OMISE_API_URL', 'https://api.omise.co/');
+define('OMISE_VAULT_URL', 'https://vault.omise.co/');
 
 class OmiseApiResource extends OmiseObject
 {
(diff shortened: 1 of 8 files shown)


includes/libraries/omise-php/lib/omise/res/OmiseApiResource.php
```diff
@@ -1,8 +1,8 @@
 <?php
 
 define('OMISE_PHP_LIB_VERSION', '2.16.0');
-define('OMISE_API_URL', 'https://api.omise.co/');
-define('OMISE_VAULT_URL', 'https://vault.omise.co/');
+define('OMISE_API_URL', 'https://api.staging-omise.co/');
```
new4762, line 4: Ah don't forgot this one 😆

new4762 (approved):

includes/libraries/omise-php/lib/omise/res/OmiseApiResource.php
aashishgurung: Oops :sweat_smile: 

Thanks.

ajzkk (approved):

ajzkk (changes requested):
It still showing the error "It seems we've been unable to process your payment properly:
customer cust_test_xxxxx was not found"

ajzkk (approved):

PR #327, opened 2022-11-01
[ENGA3-553]: Increased the delay to 2 seconds and added a check for OCBC PAO redirect URL.

aashishgurung:
#### 1. Objective

Increased the delay to 2 seconds.

Jira Ticekt: [#553](https://opn-ooo.atlassian.net/browse/ENGA3-553)

#### 2. Description of change

We observed 0.5 secs was not enough sometimes with RMS payment method while testing with Shopify. To be on safe side, we increase delay to 2 secs. Plus, we added a check for OCBC PAO redirect URL so that plugin won't identity it as user originated request because we can't add query params in OCBC PAO.

#### 3. Quality assurance

- Test with any payment method.

**🔧 Environments:**

- WooCommerce: v6.7.0
- WordPress: v6.0.2
- PHP version: 8.1
- Omise WooCommerce: 4.25.0

### DIFF
--- a/includes/class-omise-callback.php
+++ b/includes/class-omise-callback.php
@@ -38,16 +38,35 @@ class Omise_Callback {
 		$callback->validate();
 	}
 
+
+	// Sometimes cancelling a transaction does not updates the status on the Omise backend
+	// which causes the status to be pending even thought the transaction was cancelled.
+	// To avoid this random issue of status being 'Pending` when it should have been 'Cancelled',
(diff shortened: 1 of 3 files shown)


ajzkk (approved):
You can also use `sleep(2)` instead of `usleep(2000000)`

kan98 (approved):
LGTM

ajzkk (approved):
LGTM

PR #346, opened 2023-02-02
[ENGA3-443] : Resolve Sonarcloud Issue

ajzkk:
#### 1. Objective

Resolve Sonarcloud issue

[ENGA3-443](https://opn-ooo.atlassian.net/browse/ENGA3-443)

#### 2. Description of change

- Fix Sonarcloud issue
- Refactor sync order function

#### 3. Quality assurance

Checked sync order action locally for below cases

- [x] Refund fully
- [x] Refund partially
- [x] Capture
- [x] Failed
- [x] Reversed
- [x] Expired

#### 4. Impact of the change

Merchnats might not be able to sync_order manually from actions.

#### 5. Priority of change

Normal

### DIFF
--- a/includes/admin/class-omise-admin-page.php
+++ b/includes/admin/class-omise-admin-page.php
@@ -25,10 +25,6 @@ class Omise_Admin_Page extends Omise_Setting {
 			case 'error':
 				$this->errors[] = $message;
 				break;
-
-			case 'message':
-				$this->messages[] = $message;
-				break;
 			default:
 				$this->messages[] = $message;
 				break;
--- a/includes/admin/class-omise-admin-page.php
+++ b/includes/admin/class-omise-admin-page.php
@@ -29,6 +29,9 @@ class Omise_Admin_Page extends Omise_Setting {
 			case 'message':
 				$this->messages[] = $message;
 				break;
+			default:
+				$this->messages[] = $message;
+				break;
 		}
 	}
 
--- a/includes/class-omise-money.php
+++ b/includes/class-omise-money.php
@@ -51,7 +51,7 @@ class Omise_Money {
 	/**
 	 * convert omise amount to human readable amount.
 	 */
-	public static function to_readable( $amount, $currency ) {
+	public static function convert_currency_unit( $amount, $currency ) {
 		$amount   = self::purify_amount( $amount );
 		$currency = strtoupper( $currency );
 
--- a/includes/class-omise-money.php
+++ b/includes/class-omise-money.php
@@ -49,6 +49,20 @@ class Omise_Money {
 	}
 
 	/**
+	 * convert omise amount to human readable amount.
+	 */
+	public static function to_readable( $amount, $currency ) {
+		$amount   = self::purify_amount( $amount );
+		$currency = strtoupper( $currency );
+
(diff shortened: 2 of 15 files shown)


includes/gateway/abstract-omise-payment-base-card.php
```diff
@@ -232,7 +233,7 @@ public function omise_scripts() {
 		if ( is_checkout() && $this->is_available() ) {
 			wp_enqueue_script(
 				'omise-js',
-				'https://cdn.omise.co/omise.js',
+				'https://cdn.staging-omise.co/omise.js',
```
aashishgurung, line 236: Unintended change.

includes/gateway/traits/sync-order-trait.php
```diff
@@ -0,0 +1,202 @@
...
+            $order->add_order_note(sprintf($message, $e->getMessage()));
+        }
+    }
+
+    /**
+     * This function handle successful charge, 
+     * and add order not when order is partially or full refunded 
+     * when sync order action was called
+     */
+    function sync_order_handle_charge_successful($charge)
```
aashishgurung, line 65: I see lots of methods don't have visibility. SonarCloud might throw more alerts for this. If it does. then let's add access modifiers.

includes/class-omise-money.php
```diff
@@ -48,6 +48,20 @@ public static function to_subunit( $amount, $currency ) {
 		return $amount * self::$subunit_multiplier[ $currency ];
 	}
 
+	/**
+	 * convert omise amount to human readable amount.
+	 */
+	public static function to_readable( $amount, $currency ) {
```
aashishgurung, line 54: `to_readable` can imply that the amount is in some unreadable format. I would suggest to change the name `to_readable` to something like `convertDenomination` or `convertCurrencyUnit`  or `unitConversion`.

aashishgurung (approved):

includes/admin/class-omise-admin-page.php
```diff
@@ -29,6 +29,9 @@ public function add_message( $type, $message ) {
 			case 'message':
 				$this->messages[] = $message;
 				break;
+			default:
+				$this->messages[] = $message;
+				break;
```
kan98, line 34: can we remove the case `message`? or is it possible to do this? since the output is the same
```suggestion
			case 'message':
			default:
				$this->messages[] = $message;
				break;
```

includes/events/class-omise-event-charge-capture.php
```diff
@@ -78,7 +78,10 @@ public function resolve() {
 				if ( ! $this->order->has_status( 'processing' ) ) {
 					$this->order->update_status( 'processing' );
 				}
-			break;
+				break;
+
+			default:
```
kan98, line 83: should we raise an exception if we ever reach this? since something would have gone wrong right? 🤔

includes/admin/class-omise-admin-page.php
ajzkk: Yeah we can. I will remove it.

kan98 (approved):

FhanOmise (approved):

PR #353, opened 2023-03-09
Reverts - replacing credit card form with embedded form

ajzkk:
Reverts omise/omise-woocommerce#351


aashishgurung (approved):

aashishgurung:
Let's not revert the revert in master unless we have a date for the release. We might need to revert again if we need to release another version before embedded form. Let's do this in develop instead.

aashishgurung:
Done in this PR [#354](https://github.com/omise/omise-woocommerce/pull/354).

PR #360, opened 2023-03-28
Update class-omise-payment-konbini.php

bluantinoo:
Add payment link to konbini emails only if order is not paid yet

#### 1. Objective

Konbini Payment link is added in every email notification, this is confusing, both for admins that do not know what they are clicking and for customers that receive notification after successful payment.

This section will be used in the release notes. 

#### 2. Description of change

check if the order is still unpaid before adding the Konbini payment link in the email with the woocommerce_email_after_order_table action


aashishgurung:
Thank you for submitting your pull request. We have received it and We will review it as soon as possible. We appreciate your contribution to our project.

PR #365, opened 2023-04-10
[ENGA3-1134]: Installment minimum amount

aashishgurung:
#### 1. Objective

Fetch the minimum amount of Installment payment method from live config of the Opn account.

Jira Ticket: [#1134](https://opn-ooo.atlassian.net/browse/ENGA3-1134)

#### 2. Description of change

Updated the Omise PHP to use minimum amount from the capability API instead of pre-defined value. Added a function `getInstallmentMinLimits` in `includes/class-omise-capabilities.php` to get the minimum amount to show it in the checkout page.

Screenshot 2566-04-10 at 11 02 38

#### 3. Quality assurance

- Updated the `minimum installment amount` value from the Opn configuration
- Enable Installment payment method in WooCommerce.
- Add items to the cart but below the installment amount value you set up on the Opn config.
- You should see above screenshot.

- Add items to the cart above the installment amount value you set up on the Opn config.
- You should see a list of banks.

**🔧 Environments:**

- WooCommerce: v6.7.0
- WordPress: v6.0.2
- PHP version: 8.1
- Omise WooCommerce: 5.0.0

### DIFF
--- a/includes/gateway/class-omise-payment-installment.php
+++ b/includes/gateway/class-omise-payment-installment.php
@@ -79,7 +79,7 @@ class Omise_Payment_Installment extends Omise_Payment_Offsite
 			array(
 				'installment_backends' => $this->backend->get_available_providers($currency, $cart_total),
 				'is_zero_interest'     => $capabilities ? $capabilities->is_zero_interest() : false,
-				'installment_min_limit' => Omise_Money::convert_currency_unit($installmentMinLimit, $currency)
+				'installment_min_limit' => number_format(Omise_Money::convert_currency_unit($installmentMinLimit, $currency))
 			)
 		);
 	}
--- a/includes/gateway/class-omise-payment-installment.php
+++ b/includes/gateway/class-omise-payment-installment.php
@@ -72,7 +72,7 @@ class Omise_Payment_Installment extends Omise_Payment_Offsite
 		$currency   = get_woocommerce_currency();
 		$cart_total = WC()->cart->total;
 		$capabilities = $this->backend->capabilities();
-		$installmentMinLimit = $capabilities->getInstallmentLimits()['min'];
+		$installmentMinLimit = $capabilities->getInstallmentMinLimits();
 
 		Omise_Util::render_view(
 			'templates/payment/form-installment.php',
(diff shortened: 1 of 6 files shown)


ajzkk (dismissed):

includes/gateway/class-omise-payment-installment.php
```diff
@@ -72,12 +72,14 @@ public function payment_fields()
...
 		$capabilities = $this->backend->capabilities();
+		$installmentMinLimit = $capabilities->getInstallmentMinLimits();
 
 		Omise_Util::render_view(
 			'templates/payment/form-installment.php',
 			array(
 				'installment_backends' => $this->backend->get_available_providers($currency, $cart_total),
-				'is_zero_interest'     => $capabilities ? $capabilities->is_zero_interest() : false
+				'is_zero_interest'     => $capabilities ? $capabilities->is_zero_interest() : false,
+				'installment_min_limit' => number_format($installmentMinLimit / 100) // subunit to base unit
```
danfowler, line 82: @aashishgurung why don't we use this function instead?

https://github.com/omise/omise-woocommerce/blob/master/includes/class-omise-money.php#L54

aashishgurung: I forgot it existed. I have updated the code and added `number_format` in the `convert_currency_unit` since it is suppose to be human readable. Thanks.

includes/gateway/class-omise-payment-installment.php
```diff
@@ -72,12 +72,14 @@ public function payment_fields()
 		$currency   = get_woocommerce_currency();
 		$cart_total = WC()->cart->total;
 		$capabilities = $this->backend->capabilities();
+		$installmentMinLimit = $capabilities->getInstallmentMinLimits();
```
danfowler, line 75: @aashishgurung why is the function `getInstallmentMinLimits` (plural) but the variable we save it to is `$installmentMinLimit` (singular)?

includes/libraries/omise-php/lib/omise/OmiseCapabilities.php
```diff
@@ -127,7 +126,7 @@ public function makeBackendFilterChargeAmount($amount)
         return function ($backend) use ($amount, $defMin, $defMax) {
             // temporary hack for now to correct min value for installments to fixed minimum (different to normal charge minimum)
```
danfowler, line 127: @aashishgurung were we going to remove this comment as it's no longer correct?

danfowler (approved):

PR #366, opened 2023-04-10
Added a frontend validation to accept only number between 10-13 and a…

aashishgurung:
#### 1. Objective

Added a frontend validation to accept only number between 10-13 and an optional + sign in front.

Jira Ticket: [#929](https://opn-ooo.atlassian.net/browse/ENGA3-929)

#### 2. Description of change

Add a validation in the JS and fixed a bug when checkbox `same as billing address` was not selected.

#### 3. Quality assurance

Case 1:
- Adding a string should show `Phone number should be a number in Atome`

Screenshot 2566-04-10 at 18 18 23

Case 2:
- Adding no phone number should show `Phone number is required in Atome`

Screenshot 2566-04-10 at 18 18 35

Case 3
- Adding a phone number with or without `+` should redirect to Atome page.

**🔧 Environments:**

- WooCommerce: v6.7.0
- WordPress: v6.0.2
- PHP version: 8.1
- Omise WooCommerce: 5.0.0

### DIFF
--- a/assets/javascripts/omise-payment-atome.js
+++ b/assets/javascripts/omise-payment-atome.js
@@ -16,7 +16,7 @@
 		$("html, body").animate({ scrollTop:0 },"slow");
 	}
 
-    function omiseHanldeError(message) {
+    function omiseHandleError(message) {
         $form.block({
             message: null,
             overlayCSS: {
@@ -40,13 +40,13 @@
             const phoneNumber = $('#omise_atome_phone_number').val();
 			
             if (!phoneNumber) {
-                return omiseHanldeError('Phone number is required in Atome');
+                return omiseHandleError('Phone number is required in Atome');
             }
 
             const phonePattern = /(\+)?([0-9]{10,13})/;
 
             if (!phonePattern.test(phoneNumber)) {
-                return omiseHanldeError('Phone number should be a number in Atome');
+                return omiseHandleError('Phone number should be a number in Atome');
             }
 
             $form.unblock();
--- a/assets/javascripts/omise-payment-atome.js
+++ b/assets/javascripts/omise-payment-atome.js
@@ -16,25 +16,37 @@
 		$("html, body").animate({ scrollTop:0 },"slow");
 	}
 
+    function omiseHanldeError(message) {
+        $form.block({
+            message: null,
+            overlayCSS: {
+                background: '#fff url(' + wc_checkout_params.ajax_loader_url + ') no-repeat center',
+                backgroundSize: '16px 16px',
+                opacity: 0.6
+            }
+        });
+        showError(message);
+        $form.unblock();
+        return false;
+    }
+
 	function omiseFormHandler() {
         hideError();
 		if ($('#payment_method_omise_atome').is(':checked')) {
             if ($("#omise_atome_phone_default").is(":checked")) {
                 return true;
             }
+
+            const phoneNumber = $('#omise_atome_phone_number').val();
 			
-            if (!$('#omise_phone_number').val()) {
-                $form.block({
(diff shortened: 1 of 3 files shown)


assets/javascripts/omise-payment-atome.js
```diff
@@ -16,25 +16,37 @@
...
-                $form.unblock();
-                return false;
+            if (!phoneNumber) {
+                return omiseHanldeError('Phone number is required in Atome');
+            }
+
+            const phonePattern = /(\+)?([0-9]{10,13})/;
+
+            if (!phonePattern.test(phoneNumber)) {
+                return omiseHanldeError('Phone number should be a number in Atome');
```
ajzkk, line 49: Can we use `Invalid phone number` instead of `Phone number should be a number in Atome` ? Since the function did not check phone number is in Atome.

templates/payment/form-atome.php
```diff
@@ -9,7 +9,7 @@
 
         <p id="omise_atome_phone_field" class="form-row form-row-wide" style="display: none;">
             <span class="woocommerce-input-wrapper">
-                <input id="omise_phone_number" class="input-text" name="omise_phone_number" type="tel" autocomplete="off">
+                <input id="omise_atome_phone_number" class="input-text" name="omise_atome_phone_number" type="tel" autocomplete="off">
```
ajzkk, line 12: Can we set initial value of Atome phone number from billing phone number?
Since there is no label, can we also add placeholder?

assets/javascripts/omise-payment-atome.js
aashishgurung: The function retrieves the phone number from Atome and `Invalid phone number` would confuse user with billing phone number.

ajzkk (approved):

assets/javascripts/omise-payment-atome.js
danfowler, line 49:
```suggestion
                return omiseHandleError('Phone number should be a number in Atome');
```

assets/javascripts/omise-payment-atome.js
```diff
@@ -16,25 +16,37 @@
...
-                        background: '#fff url(' + wc_checkout_params.ajax_loader_url + ') no-repeat center',
-                        backgroundSize: '16px 16px',
-                        opacity: 0.6
-                    }
-                });
-                showError('Phone number is required in Atome');
-                $form.unblock();
-                return false;
+            if (!phoneNumber) {
+                return omiseHanldeError('Phone number is required in Atome');
```
danfowler, line 43:
```suggestion
                return omiseHandleError('Phone number is required in Atome');
```

includes/gateway/class-omise-payment-atome.php
```diff
@@ -145,10 +145,12 @@ private function validateMinRequiredAmount()
     public function charge($order_id, $order)
     {
         $currency = $order->get_currency();
-        $default_phone_selected = $_POST['omise_atome_phone_default'];
-        $phone_number = isset($default_phone_selected) && 1 == $default_phone_selected ?
+        $default_phone_selected = isset($_POST['omise_atome_phone_default']) ?
+            $_POST['omise_atome_phone_default']
+            : false;
+        $phone_number = true === (bool)$default_phone_selected ?
```
danfowler, line 151: Why do we need to compare with `true` if `(bool)` casts `$default_phone_selected` to a boolean?

aashishgurung: We don't need to compare. Thanks for noticing it. I have made the changes.

danfowler (approved):

PR #372, opened 2023-04-26
Added  PayPay

ajzkk:
#### 1. Objective

Added  PayPay

#### 2. Description of change

Screenshot 2023-04-26 at 3 24 33 PM

#### 3. Quality assurance

- Switch currency to `JPY` in WooCommerce setting
- Enter Japan PSP credentials and Enable PayPay payment method

#### 4. Impact of the change

N/A

#### 5. Priority of change

Normal


includes/gateway/class-omise-payment-paypay.php
```diff
@@ -0,0 +1,86 @@
...
+	 */
+	public function charge($order_id, $order)
+	{
+		$currency = $order->get_currency();
+		return OmiseCharge::create([
+			'amount' => Omise_Money::to_subunit($order->get_total(), $currency),
+			'currency' => $currency,
+			'description' => apply_filters('omise_charge_params_description', 'WooCommerce Order id ' . $order_id, $order),
+			'source' => ['type' => $this->source_type],
+			'return_uri' => $this->getRedirectUrl('omise_paypay_callback', $order_id, $order),
```
danfowler, line 66: @ajzkk when can we start refactoring some of these repeated patterns/strings? e.g.

https://github.com/omise/omise-woocommerce/blob/bc6bbbfc6ea7edfa5176c999472bc5df129e987b/includes/gateway/class-omise-payment-alipayplus.php#L88

includes/gateway/class-omise-payment-paypay.php
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function charge($order_id, $order)
+	{
+		$currency = $order->get_currency();
+		return OmiseCharge::create([
+			'amount' => Omise_Money::to_subunit($order->get_total(), $currency),
```
annatotest, line 62: But Japanese Yen does not have subunits.

danfowler: @annatotest Good catch @annatotest. `.to_subunit` explicitly treats subunits as equal to units for `JPY` 

https://github.com/omise/omise-woocommerce/blob/bc6bbbfc6ea7edfa5176c999472bc5df129e987b/includes/class-omise-money.php#L11-L20

If it were the case that this payment method accepted other currencies, this code would still work.

includes/gateway/class-omise-payment-paypay.php
ajzkk: Shall we create a ticket under 2023Q2 - PHP modernization road map for that?

includes/gateway/class-omise-payment-paypay.php
ajzkk: Yeah, no need to worry. I have tested with JPY currency on test mode, it work well.

includes/gateway/class-omise-payment-paypay.php
danfowler: It seems distinct to me. PHP modernization is more about adopting modern approaches to working with PHP and integrating with other frameworks. My request above is specific to how we implemented in WooCommerce.

danfowler (approved):

assets/images/paypay.png
aashishgurung, line 1: Can we please reduce the PayPay image size? It's too big than what is required.

ajzkk: Updated `1.58 KB`

ajzkk: I can convert from svg to png, which will reduce from `115 KB` to `3 KB`. 
Shall we do in this same PR?

danfowler: Yes, let's do.

danfowler: @ajzkk Something must be weird with the original SVG to make it so large in the first place. SVG could be the smallest possible format.

aashishgurung: In the future, if resize the image to the height & width we require then I think we could bring down the image even more. Maybe under 1KB. We can talk with designers about the size we need. They can reduce the height and width along with the size of the Firstchoice while maintaining the file format as SVG.

ajzkk: @aashishgurung  Setting height and width on SVG doesn't change file size.
@danfowler original SVG have large size because it have multiple layers of shadow effect on it. 
But Not sure where this logo was originally get from.

aashishgurung: Current, PayPay is in PNG format. So, I thought maybe we can achieve more. But, we can end it here. The current version looks good.

aashishgurung (approved):

PR #374, opened 2023-05-10
Bug fixed on guest user checkout

ajzkk:
#### 1. Objective

Bug fixed on guest user checkout

#### 2. Description of change

When `$user = $order->get_user();` is empty, `$user->test_omise_customer_id`  or `$user->live_omise_customer_id` will throw `attempt to read property test_omise_customer_id from null`.

So we need to check user is not empty before we get value from user.

#### 3. Quality assurance

Create manual order with guest user from admin panel and use customer payment page link to make payment where user is  not logged in (eg. try with different browser).

#### 4. Impact of the change

N/A

#### 5. Priority of change

Normal

### DIFF
--- a/includes/gateway/abstract-omise-payment-base-card.php
+++ b/includes/gateway/abstract-omise-payment-base-card.php
@@ -24,11 +24,7 @@ abstract class Omise_Payment_Base_Card extends Omise_Payment
 		}
 
 		$user = $order->get_user();
-		$omise_customer_id = null;
-
-		if(!empty($user)) {
-			$omise_customer_id = $this->is_test() ? $user->test_omise_customer_id : $user->live_omise_customer_id;
-		}
+		$omise_customer_id = $this->getOmiseCustomerId($user);
 
 		// Saving card.
 		$saveCustomerCard = $_POST['omise_save_customer_card'];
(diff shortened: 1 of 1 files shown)


annatotest (approved):

includes/gateway/abstract-omise-payment-base-card.php
```diff
@@ -24,7 +24,11 @@ public function charge($order_id, $order)
 		}
 
 		$user = $order->get_user();
-		$omise_customer_id = $this->is_test() ? $user->test_omise_customer_id : $user->live_omise_customer_id;
+		$omise_customer_id = null;
+
```
aashishgurung, line 28: Let's start breaking functions into smaller unit when it makes sense. Here, I see we can move the responsibility of getting customer ID to a different method thus adhering to `separation of concerns` and `single responsibility principle`. Plus, if we had to make any change to get the customer ID in the future then we do not need to touch `charge` function adhering to `open-close principle`.


```
$omise_customer_id = $this->getCustomerId($user)
...
...
private function getCustomerId($user)
{
	if(!empty($user)) {
		return null;
	}

	return $this->is_test() ? $user->test_omise_customer_id : $user->live_omise_customer_id;
}
```

ajzkk: Fixed.

annatotest (approved):

aashishgurung (approved):

PR #378, opened 2023-05-19
[ENGA3-1334]: Remove free products from the Atome bill

aashishgurung:
#### 1. Objective

Remove free or complimentary products from the Atome bill

Jira Ticket: [#1334](https://opn-ooo.atlassian.net/browse/MIT-1334)

#### 2. Description of change

Add a logic to check the price of subtotal which is the total amount of the products combined excluding shipping cost.

Screenshot 2566-05-19 at 12 30 26

#### 3. Quality assurance

- Add a freebie product in the cart
- Try to pay via Atome

**🔧 Environments:**

- WooCommerce: v6.7.0
- WordPress: v6.0.2
- PHP version: 8.1
- Omise WooCommerce: 5.1.0

### DIFF
--- a/includes/gateway/class-omise-payment-atome.php
+++ b/includes/gateway/class-omise-payment-atome.php
@@ -80,12 +80,12 @@ class Omise_Payment_Atome extends Omise_Payment_Offsite
     public function payment_fields()
     {
         parent::payment_fields();
-        $viewData = $this->validateMinRequiredAmount();
+        $viewData = $this->validateAtomeRequest();
 
         Omise_Util::render_view('templates/payment/form-atome.php', $viewData);
     }
 
-    private function validateMinRequiredAmount()
+    private function validateAtomeRequest()
     {
         $limits = [
             'thb' => [
@@ -103,7 +103,14 @@ class Omise_Payment_Atome extends Omise_Payment_Offsite
         ];
 
         $currency = strtolower(get_woocommerce_currency());
-        $cartTotal = WC()->cart->total;
+        $cart = WC()->cart;
+
+        if ($cart->subtotal === 0) {
+            return [
+                'status' => false,
+                'message' => 'Complimentary products cannot be billed.'
+            ];
+        }
 
         if (!isset($limits[$currency])) {
             return [
@@ -114,7 +121,7 @@ class Omise_Payment_Atome extends Omise_Payment_Offsite
 
         $limit = $limits[$currency];
 
-        if ($cartTotal < $limit['min']) {
+        if ($cart->total < $limit['min']) {
             return [
                 'status' => false,
                 'message' => sprintf(
@@ -125,7 +132,7 @@ class Omise_Payment_Atome extends Omise_Payment_Offsite
             ];
         }
 
-        if ($cartTotal > $limit['max']) {
+        if ($cart->total > $limit['max']) {
             return [
                 'status' => false,
                 'message' => __(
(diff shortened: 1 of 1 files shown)


includes/gateway/class-omise-payment-atome.php
```diff
@@ -191,14 +198,20 @@ private function getItems($order, $currency)
 
         // Loop through ordered items
         foreach ($items as $key => $item) {
+            // item don't have price. So we have to take subtotal and divide it by quantity to get the price
+            $pricePerItem = $item['subtotal'] / $item['qty'];
+
+            // Remove product from the list if the price is 0
+            if ((float)$pricePerItem === 0.00) {
```
danfowler, line 205: @aashishgurung mathematically, the only way for `(float)$pricePerItem` to be `0.00` is for `$item['subtotal']` to be `0` 🤔  so couldn't we leave `$pricePerItem` where it was originally defined and use `$item['subtotal']` here?

includes/gateway/class-omise-payment-atome.php
```diff
@@ -202,7 +202,7 @@ private function getItems($order, $currency)
             $pricePerItem = $item['subtotal'] / $item['qty'];
 
             // Remove product from the list if the price is 0
-            if ((float)$pricePerItem === 0.00) {
+            if ((float)$item['subtotal'] === 0.00) {
```
danfowler, line 205: Why do we need to cast to `float`?

aashishgurung: So that the amount between 0 and 1 would be represented correctly. The minimum price can be satisfied by the shipping cost. This might not happen in real life but just in case.

ajzkk: Since you are casting as float.
If amount is RM 0.8 then casting to float will not give 0, it will be still 0.8.
are you trying to do `(int)$item['subtotal'] === 0`?

includes/gateway/class-omise-payment-atome.php
```diff
@@ -103,7 +103,14 @@ private function validateMinRequiredAmount()
         ];
 
         $currency = strtolower(get_woocommerce_currency());
-        $cartTotal = WC()->cart->total;
+        $cart = WC()->cart;
+
+        if ($cart->subtotal === 0) {
```
ajzkk, line 108: Just curious, why subtotal value zero mean complimentary products?

includes/gateway/class-omise-payment-atome.php
aashishgurung: Yes since it will give 0.8, it means 0.8 is not equal to 0.00. So, we add the item to the product list we send to Atome. If we cast it to integer then it will treat non-zero (0.8) as zero.

includes/gateway/class-omise-payment-atome.php
aashishgurung: here, complimentary products mean free products. A product for which customer don't have to pay. This does not include shipping cost.

ajzkk: Is there any chance `$cart->subtotal` value in float, mean `0.0`. 
Since we are using `===` which mean `0.0 != 0`

aashishgurung: WooCommerce will send the subtotal as 0, it won't be 0.0. The only case where we get the decimal number is if the amount itself is in float or if we do some calculation.

ajzkk (approved):

danfowler (approved):

PR #379, opened 2023-06-01
Bug Fixed not compatible with country based payments plugin

ajzkk:
#### 1. Objective

Bug Fixed not compatible with country based payments plugin

#### 2. Description of change

Country based payments plugin remove the the whole html element of omise card payment, if user not selected the country in shipping address.

#### 3. Quality assurance

Tested locally with 
- For first time user 
- Tested with normal user
- Tested with guest user

#### 4. Impact of the change

N/A

#### 5. Priority of change

Normal

### DIFF
--- a/assets/javascripts/omise-payment-form-handler.js
+++ b/assets/javascripts/omise-payment-form-handler.js
@@ -255,9 +255,10 @@
 	}
 
 	function initializeSecureCardForm() {
-		if (Boolean(omise_params.secure_form_enabled)) {
+		const omiseCardElement = document.getElementById('omise-card');
+		if (omiseCardElement && Boolean(omise_params.secure_form_enabled)) {
 			showOmiseEmbeddedCardForm({
-				element: document.getElementById('omise-card'),
+				element: omiseCardElement,
 				publicKey: omise_params.key,
 				hideRememberCard: HIDE_REMEMBER_CARD,
 				locale: LOCALE,


assets/javascripts/omise-payment-form-handler.js
```diff
@@ -255,9 +255,10 @@
 	}
 
 	function initializeSecureCardForm() {
-		if (Boolean(omise_params.secure_form_enabled)) {
+		const omiseCardElement = document.getElementById('omise-card');
```
danfowler, line 258: How do we test?

aashishgurung:
Can you add more description about how other plugins removed code from our plugin?

assets/javascripts/omise-payment-form-handler.js
ajzkk: I can recreate the issue locally by following the flow described by CS team.

danfowler: But what about unit tests?

ajzkk: Not possible to write unit test currently. It is not based on npm.

ajzkk:
> Can you add more description about how other plugins removed code from our plugin? How does this change prevents that?

I do not know how their plugin work, but I can visually see in checkout page that card payment is missing, and also if you try to inspect, you will see card payment element is missing until user select the country. So I check with the element. If element is missing, then do not run `showOmiseEmbeddedCardForm`

danfowler (approved):

aashishgurung (approved):

PR #388, opened 2023-08-04
Fix installment payment when admin manually pay for order

aashishgurung:
#### 1. Objective

Fix installment payment when admin manually pay for order.

#### 2. Description of change

Installment payment was not functioning properly when admin manually pay for order because the total amount was 0. This happened because logic was getting the total from the cart. But, cart is not present when admin pay for an order. So, we added a check that will get the total amount from order if the payment page is triggered by admin. If not then we get the total from cart.

Screenshot 2566-08-04 at 09 43 19
<br/><br/>

#### 3. Quality assurance

Specify where and how you tested this and what further testing it might need.

**🔧 Environments:**

- WooCommerce: v7.8.2
- WordPress: v6.2.2
- PHP version: 8.1
- Omise WooCommerce: 5.1.1

### DIFF
--- a/assets/javascripts/omise-embedded-card.js
+++ b/assets/javascripts/omise-embedded-card.js
@@ -25,11 +25,17 @@ function showOmiseEmbeddedCardForm({
   }
   element.style.height = iframeElementHeight + 'px'
 
+  const localeMatching = {
+    en_US: 'en',
+    ja_JP: 'ja',
+    th_TH: 'th'
+  }
+
   OmiseCard.configure({
     publicKey: publicKey,
     element,
     customCardForm: true,
-    locale: locale,
+    locale: localeMatching[locale] ?? 'en',
     customCardFormTheme: theme,
     customCardFormHideRememberCard: hideRememberCard ?? false,
     customCardFormBrandIcons: brandIcons ?? null,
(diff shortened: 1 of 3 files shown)


assets/javascripts/omise-embedded-card.js
```diff
@@ -25,17 +25,11 @@ function showOmiseEmbeddedCardForm({
   }
   element.style.height = iframeElementHeight + 'px'
 
-  const localeMatching = {
```
danfowler, line 28: Since this is a distinct fix, can we implement in separate PR?

aashishgurung: Sure. It's better to do it in separate PR.

includes/gateway/class-omise-payment-installment.php
```diff
@@ -84,6 +85,26 @@ public function payment_fields()
 		);
 	}
 
+	/**
+	 * Get the total amount of an order
+	 */
+	public function getTotalAmount()
```
ajzkk, line 91: Can we move this function under `Omise_Payment_Offsite` class to reuse if we needed for other payment methods in the future.

aashishgurung: We could but we don't enough information on this now. We are not sure if it will be applicable for offsite payments only. It might be needed for offline payment as well. In that case, `Omise_Payment` would be appropriate. Or we might never need it outside installment. So, I would suggest to do it when needed.

ajzkk (approved):

danfowler (approved):

PR #408, opened 2023-10-12
Add QR expires count down for promptpay

ajzkk:
#### 1. Objective

Add QR expires count down for promptpay

https://opn-ooo.atlassian.net/browse/MIT-1889

#### 2. Description of change

- Show QR code expire count down instead of date time.
- Created `get_pending_status` function since it does not support to mock constant.
- Created `file_get_contents` wrapper class since it does not support to mock global function.

Screenshot 2023-10-12 at 1 44 09 PM

#### 3. Quality assurance

- Crate payment with promptpay QR  and checked count down is correct.
- And wait for expiration and refresh the page which should be hidden QR code image.

#### 4. Impact of the change

N/A

#### 5. Priority of change

Normal

### DIFF
--- a/assets/javascripts/omise-promptpay-count-down.js
+++ b/assets/javascripts/omise-promptpay-count-down.js
@@ -1,0 +1,43 @@
+(function () {
+    let countDownInterval;
+
+    function calculateCountdown() {
+        const currentDateTime = new Date();
+        const expiresAtDateTime = new Date(omise.qr_expires_at);
+        const difference = expiresAtDateTime - currentDateTime;
+        const hours = Math.floor((difference % (1000 * 60 * 60 * 24)) / (1000 * 60 * 60));
+        const minutes = Math.floor((difference % (1000 * 60 * 60)) / (1000 * 60));
+        const seconds = Math.floor((difference % (1000 * 60)) / 1000);
+        return { hours, minutes, seconds };
+    }
+
+    function padZero(num) {
+        return num.toString().padStart(2, '0');
+    }
+
+    function updateCountdown(fromInterval = true) {
+        const countdownDisplay = document.getElementById('countdown');
+        if(!countdownDisplay) {
+            return;
+        }
+
+        const { hours, minutes, seconds } = calculateCountdown();
+
+        if (hours + minutes + seconds < 0) {
+            // To prevent infinite loading, we need to reload and clear interval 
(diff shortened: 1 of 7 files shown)


includes/gateway/class-omise-payment.php
```diff
@@ -132,6 +132,26 @@ public function protectMetadata($protected, $metadataKeys)
...
+
+	/**
+	 * file_get_contents wrapper
+	 * 
+	 * Since we cannot mock global function,
+	 * we have to create a wrapper for file_get_contents.
+	 */
+	public function file_get_contents( $url ) {
+		return file_get_contents( $url );
+	}
```
aashishgurung, line 153: Can we move these to their own helper functions?

assets/javascripts/omise-promptpay-count-down.js
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        const { hours, minutes, seconds } = calculateCountdown();
+
+        if (hours + minutes + seconds < 0) {
+            // To prevent infinite loading, we need to reload and clear interval 
+            // only when it is from setInterval function.
+            if (fromInterval) {
+                clearInterval(countDownInterval)
+                window.location.reload()
```
aashishgurung, line 31: Instead of reload, can we simply remove the counter and add expiration alert?

aashishgurung: Can you tell me more about `To prevent infinite loading, we need to reload and clear interval`?

ajzkk: To prevent infinite loading, we need to reload and clear interval only when it is from setInterval function.

aashishgurung: It appears we may not have fully considered all aspects of this. Perhaps we can enhance it in a future ticket.

includes/gateway/class-omise-payment.php
ajzkk: I have moved `file_get_contents` to its own helper class. 
For `get_pending_status`, I think it still can be under  `Omise_Payment` since it is to get it's own constant.

aashishgurung (dismissed):
Let's remove `file_get_contents` from Omise_Payment. Other than that LGTM.

AnasNaouchi (approved):

PR #427, opened 2023-12-18
Fix auto scroll to secure form issue

aashishgurung:
#### 1. Objective

Fix auto scroll to secure form issue

#### 2. Description of change

- Added a check on existence of the secure form on `updated_checkout` event. This event was triggering when we update postal code.
- Hidden the secure form when iframe is loaded to avoid auto scroll to the iframe. We make is visible again after 1 sec.

#### 3. Quality assurance

Precondition:
- Use mobile device
- Enable secure form from WP admin
- There should not be any saved card

Case 1:
- Add items to cart and checkout
- Fill in the billing info.
- When you update postal code, the focus should not be shifted to credit card form.

Case 2:
- Add items to cart and checkout with credit card. This will set default payment to credit card on your next order
- For the next order, add items to cart and go to checkout page. The page should not scroll down to the credit card form.

**🔧 Environments:**

- WooCommerce: v8.3.1
- WordPress: v6.4.2
- PHP version: 8.1

#### 4. Impact of the change

- N/A

#### 5. Priority of change

Normal

#### 6. Additional Notes

- N/A

### DIFF
--- a/assets/javascripts/omise-embedded-card.js
+++ b/assets/javascripts/omise-embedded-card.js
@@ -66,4 +66,20 @@ function showOmiseEmbeddedCardForm({
     onCreateTokenSuccess: onSuccess ?? noop,
     onError: onError ?? noop
   });
+
+  preventAutoScrollToSecureForm()
+}
+
+// Link: https://stackoverflow.com/questions/6596668/iframe-on-the-page-bottom-avoid-automatic-scroll-of-the-page
+function preventAutoScrollToSecureForm() {
+  const iframe = document.getElementById('omise-checkout-iframe-app');
+
+  iframe.addEventListener('load', function () {
+    // hide iframe to avoid auto scroll to secure form
+    iframe.style.display = "none";
+
+    // Bring back secure form
+    // why 1000? 0-800: didn't work, 900: Gave flaky results
+    setTimeout(() => iframe.style.display = "block", 1000)
+  });
 }
--- a/assets/javascripts/omise-payment-form-handler.js
+++ b/assets/javascripts/omise-payment-form-handler.js
@@ -295,7 +295,12 @@
 		});
 
 		$(document).on('updated_checkout', function () {
-			initializeSecureCardForm();
+			const iframe = document.getElementById('omise-checkout-iframe-app');
+
+			// Ignore if secure form is already initialized.
+			if (!iframe) {
+				initializeSecureCardForm();
+			}
 		});
 
 		initializeSecureCardForm();


AnasNaouchi (approved):

assets/javascripts/omise-embedded-card.js
```diff
@@ -66,4 +66,20 @@ function showOmiseEmbeddedCardForm({
...
+function preventAutoScrollToSecureForm() {
+  const iframe = document.getElementById('omise-checkout-iframe-app');
+
+  iframe.addEventListener('load', function () {
+    // hide iframe to avoid auto scroll to secure form
+    iframe.style.display = "none";
+
+    // Bring back secure form
+    // why 1000? 0-800: didn't work, 900: Gave flaky results
+    setTimeout(() => iframe.style.display = "block", 1000)
```
danfowler, line 83: Instead of relying on a timeout, is there any event we can hook this on?

assets/javascripts/omise-embedded-card.js
```diff
@@ -66,4 +66,20 @@ function showOmiseEmbeddedCardForm({
     onCreateTokenSuccess: onSuccess ?? noop,
     onError: onError ?? noop
   });
+
+  preventAutoScrollToSecureForm()
+}
+
+// Link: https://stackoverflow.com/questions/6596668/iframe-on-the-page-bottom-avoid-automatic-scroll-of-the-page
+function preventAutoScrollToSecureForm() {
```
danfowler, line 74: * Will this need to be propagated across all users of the secure card embedded form? e.g. Magento
* Is there any room to move the responsibility for this to pay.js itself?

assets/javascripts/omise-embedded-card.js
aashishgurung: As per what we are trying to achieve which is to run this code after an iframe is loaded but with a delay, I couldn't fin any event that satisfy our use case.

assets/javascripts/omise-embedded-card.js
aashishgurung: - Will this need to be propagated across all users of the secure card embedded form? e.g. Magento
   - I haven't tested it in Magento but it's safe to assume yes.
- Is there any room to move the responsibility for this to pay.js itself?
   - It's OmiseJS creating the iframe, not PayJS but I did the same logic there first. It didn't work.
   
  The reasons on why it didn't work could be that
  - we are attaching the iframe to an element of the parent page (WooCommerce in this case).
  - the iframe's code originates from a different domain than the parent page.

These restrictions might have prevented it from accessing and manipulating elements outside the iframe's boundaries.

danfowler: What I mean is, can the form element of pay.js that is stealing focus be made to not do so?

aashishgurung:
Closing this PR as the issue can be resolved by making changes in the dependent JS pacakge.

PR #437, opened 2024-03-05
Library warning

muthuswamyopn:
Added security warning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Opn Payments provides a set of APIs that help merchants of any size accept payme
 
 ## Security Warning
 
-**Do NOT use versions less than 4.8 as they are outdated and have security vulnerabilities.**
+**Do NOT use any version less than 4.8 as they are outdated and have security vulnerabilities.**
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,6 +7,12 @@ Opn Payments provides a set of APIs that help merchants of any size accept payme
 
 **Omise WooCommerce** is our official plugin providing support for processing payments on WooCommerce through Opn Payments.
 
+## Security Warning
+
+**Please do NOT use versions less than 4.8, as they are outdated and have security vulnerabilities..**
```
aashishgurung, line 12: @muthuswamyopn I think it would be better to include sub-string `Omise WooCommerce` before `4.8` so that readers would be clear that we are talking about plugin and not the WordPress or WooCommerce.

aashishgurung, line 12:
```suggestion
**Please do NOT use versions less than 4.8, as they are outdated and have security vulnerabilities.**
```

aashishgurung (approved):

PR #462, opened 2024-05-31
[MIT-1912] Implement whitelabel installment

pynkittisak:
## Description
- Added whitelabel installment for woocommerce
image
*NOTE:*

## Rollback procedure

Revert PR

### DIFF
--- a/includes/gateway/class-omise-payment-installment.php
+++ b/includes/gateway/class-omise-payment-installment.php
@@ -119,13 +119,11 @@ class Omise_Payment_Installment extends Omise_Payment_Offsite
 	 */
 	public function charge($order_id, $order)
 	{
-		$token = isset( $_POST['omise_token'] ) ? wc_clean( $_POST['omise_token'] ) : '';
-		$source = isset( $_POST['omise_source'] ) ? wc_clean( $_POST['omise_source'] ) : '';
-		$requestData = $this->get_charge_request($order_id, $order, $token, $source);
+		$requestData = $this->get_charge_request($order_id, $order);
 		return OmiseCharge::create($requestData);
 	}
 
-	public function get_charge_request($order_id, $order, $token, $source)
+	public function get_charge_request($order_id, $order)
 	{
 		$source_type = $_POST['source']['type'];
 		$source_type = isset($source_type) ? $source_type : '';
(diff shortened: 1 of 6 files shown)


AnasNaouchi (dismissed):

templates/payment/form-installment.php
```diff
@@ -1,49 +1,9 @@
...
-			<p>
-				<?php echo $viewData['is_zero_interest'] ? __( 'All installment payments are interest free', 'omise' ) : __( 'Monthly payment rates shown may be inaccurate as interest rates are subject to change by its bank issuer.', 'omise' ); ?>
-			</p>
-		</div>
-	</fieldset>
+	<div id="omise-installment" style="width:100%; max-width: 400px;"></div>
+	<script>
+		window.LOCALE = `<?php echo get_locale(); ?>`;
+		window.OMISE_CUSTOM_FONT_OTHER = 'Other';
+	</script>
```
aashishgurung, line 6: By removing the old form, the current installment payment will not be visible. We still need to support the old installment.

includes/gateway/class-omise-payment-installment.php
```diff
@@ -105,16 +106,26 @@ public function getTotalAmount()
...
+	}
+
 	/**
 	 * @inheritdoc
 	 */
 	public function charge($order_id, $order)
 	{
-		$requestData = $this->get_charge_request($order_id, $order);
+		$token = isset( $_POST['omise_token'] ) ? wc_clean( $_POST['omise_token'] ) : '';
+		$source = isset( $_POST['omise_source'] ) ? wc_clean( $_POST['omise_source'] ) : '';
```
aashishgurung, line 123: Let's move this to `get_charge_request` and we don't have to pass these.

templates/payment/form-installment.php
pynkittisak: Hmm, But it will be rendered by pay.js, which we have decided to use either WLB or non-WLB internally.
For account that don't have whitelabel installment enabled, will see the old installment with the new ui.

includes/gateway/class-omise-payment-installment.php
pynkittisak: Nice.

templates/payment/form-installment.php
aashishgurung: Ah okay. I didn't realize that.

aashishgurung: Ah okay. That will work.

AnasNaouchi (approved):

PR #518, opened 2025-04-03
[MIT-3307][WooCommerce] Mobile banking payment method is failing

rosle:
## Description

Update the view that is not accessing deprecated payment method fields to use the new one.

This bugs happens after we updated the plugin to use new Omise API `2019-05-29`.

### More information

Previously we're using payment backends`_id` which is built from the key returned in [Capability: payment_backends](https://docs.omise.co/capability-api?version=2017-11-02) of the API `2017-11-02`. The field is being used in the mobile banking payment method view, causing the the mobile bank to to break (Cannot select the bank)

In the latest API version, we can use `name` in [Capability: payment_methods](https://docs.omise.co/capability-api?version=2019-05-29). So I updated the view from using `_id` to `name`.

I also found another place that is using old `_id`, which is `touch_n_go`. I already put a fix here as well.

### Related links:

- https://opn-ooo.atlassian.net/browse/MIT-3307

## Rollback procedure

`default rollback procedure`

### DIFF
--- a/includes/gateway/class-omise-payment.php
+++ b/includes/gateway/class-omise-payment.php
@@ -586,7 +586,7 @@ abstract class Omise_Payment extends WC_Payment_Gateway {
 
         $index = array_search($this->source_type, array_column($this->payment_settings['backends'], 'name'));
 
-        if (!$index) {
+        if ($index === false) {
             return null;
         }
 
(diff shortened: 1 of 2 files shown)


includes/gateway/class-omise-payment.php
```diff
@@ -584,9 +584,9 @@ public function get_provider()
             return null;
         }
 
-        $index = array_search($this->source_type, array_column($this->payment_settings['backends'], '_id'));
+        $index = array_search($this->source_type, array_column($this->payment_settings['backends'], 'name'));
 
-        if (!$index) {
+        if ($index === false) {
```
rosle, line 589: **note:** ‏Update the condition since `$index` can return `0` when the search item is the first element.

FhanOmise (approved):

Narum11 (approved):

PR #522, opened 2025-04-30
[MIT-3125] Unable to make installment payment with WooCommerce Block

rosle:
## Description

- Fix installment payment form doesn't show up.
- Add pre-commit hook to build the JS assets to make sure build files are up-to-date.

#### Fix installment payment form doesn't show up

The issue comes from when accessing `window.wc.wcBlocksData.CART_STORE_KEY`, it returns null.
So we cannot get the cart info from WooCommerce. It prevents the form to be shown.

🛠️ To fix, we need to properly declare the script dependencies to make sure `wcBlocksData` is available. For that I've adjusted the webpack config to include `wcBlocksData`. Then when using `@woocommerce/block-data`, the dependencies are generated correctly.

|Before|After|
|---|---|
|Screenshot 2025-04-30 at 6 12 19 PM|Screenshot 2025-04-30 at 6 08 38 PM|

### Related links:

https://opn-ooo.atlassian.net/browse/MIT-3125

## Rollback procedure

`default rollback procedure`


Narum11 (approved):

FhanOmise (approved):

tookaa2:
approved

